Issue #4, opened 2016-08-12
Allow the tool to be used in either a Python 2 or Python 3 environment


konklone:
This means replacing the use of `sslyze` with something that can work in Python 3, or separating the `sslyze` component out in some way (by having the user rely on Docker or pyenv or something). 

For what the tool currently does (analysis of HTTP behavior and HTTPS failure behavior), I think `sslyze` is wholly replace-able. `sslyze`, or something like it, would be more necessary for analyzing other TLS qualities, like use of ciphers or protocol versions.

Replacing `sslyze` would mean digging into `requests`' API more and analyzing certificate validation errors. This would also have the benefit of speeding up use of the tool in development considerably, because these would be included in the transparent `requests` caching added in #3.

However, we do want to still allow the tool to be used in a Python 2 environment, if at all possible. I believe that's doable.

garrettr:
> This means replacing the use of sslyze with something that can work in Python 3

In an ideal world, this issue would be resolved by making sslyze compatible with Python 3. This would enable us to make pshtt compatible with Python 2.x and 3.x, while maintaining the potential to support future types of scans which are only possible via sslyze.

This is not impossible: the biggest hurdle appears to be porting nassl, the Python C extension that implements the OpenSSL bindings required by sslyze, to be compatible with Python 3. An external contributor has volunteered to do this, and appears to be making progress on their [branch](https://github.com/siteblindado/nassl/tree/python3). I just commented on the relevant upstream issue with the intention of reviving it: https://github.com/nabla-c0d3/sslyze/issues/138.

garrettr:
As stated in the issue description, the easiest way to enable Python 3 support is to remove the dependency on sslyze. To do so, we need to determine if we can at least achieve feature parity between sslyze and another library (preferably `requests`, since it's already being used in pshtt) for the scan results that are currently generated by sslyze.

pshtt derives the following results from sslyze:

* `https_expired_cert`
* `https_bad_chain`
* `https_bad_hostname`

`requests` responds to any SSL errors during a request by raising `SSLError`. This is not as explicit as I would like it to be, but a quick survey of the requests code shows that context-specific information is typically passed to the SSLError constructor, albeit in a cornucopia of types/formats.

Let's see what exceptions are raised by `requests.get` for examples of each of the above types of errors:

```py
import requests

urls = [
    'expired.badssl.com',
    'untrusted-root.badssl.com',
    'wrong.host.badssl.com'
]

for url in urls:
    try:
[...]

PR #7, opened 2016-08-16
Refactor domain/endpoint classes, move HSTS parsing out of sslyze

konklone:
This does two main things:

* Refactors `port80` and `port443` into one `Endpoint` class that can work for any of the 4 endpoints, and puts them in a parent `Domain` class. Right now, the `Domain` class is a dumb container, and all 4 endpoints are passed around in method arguments anyway. However, this container makes semantic sense, and provides a logical place for future methods that speak to all endpoints.

* Has the HSTS parser parse the HSTS header from the original `requests` hit, rather than the output of SSLyze. There wasn't anything special about `sslyze` here, it was just used to get the header string. This gets us more out of the `requests` hits, and potentially prepares us for #4 (removing `sslyze` and making this Python version agnostic).


h-m-f-t:
All systems go! Thanks for hitting the README, too.

konklone:
I should have mentioned this, but for simplicity's sake, in this PR I also removed the `Broken Root`, `Broken WWW`, and `Weak Signature` fields, as they are not used by any client I'm aware of right now, either computer or human. If that's incorrect, I can easily add them back in.

Issue #8, opened 2016-08-16
Import public suffix list


konklone:
Not a high priority, but noting for the future (and to make this tool more general purpose) that the idea of "parent domain" and "subdomain" depends on the [Public Suffix List](https://publicsuffix.org/), and we should download/cache it like we do the preload list.

Python library to do this: https://pypi.python.org/pypi/publicsuffix/

konklone:
Looks like GitHub's auto-closer caught "resolve #8" in #36 as meaning this should be auto-closed. Re-opening.

garrettr:
Some motivation to fix this: digital.newint.com.au would like to be added to Secure the News, but pshtt cannot scan their site because they use the `.com.au` eTLD.

konklone:
@garrettr @sighmon Could you be more specific about what's not working? `pshtt` can scan the site, it's the definition of "parent domain" used to calculate some HSTS/preloading stuff that would be affected. 

Below is a scan result for `digital.newint.com.au` I just ran. It's possible there's a logic bug in there relating to this, but if you can identify what it is it'll help us prioritize the work, and form a test case for concluding that it's been fixed:

```json
[
  {
    "Base Domain": "com.au", 
    "Canonical URL": "https://digital.newint.com.au", 
    "Defaults to HTTPS": true, 
    "Domain": "digital.newint.com.au", 
    "Domain Enforces HTTPS": true, 
    "Domain Supports HTTPS": true, 
    "Domain Uses Strong HSTS": true, 
    "Downgrades HTTPS": false, 
    "HSTS": true, 
    "HSTS Entire Domain": true, 
    "HSTS Header": "max-age=631152000; includeSubdomains; preload", 
    "HSTS Max Age": 631152000, 
    "HSTS Preload Pending": false, 
    "HSTS Preload Ready": true, 
    "HSTS Preloaded": false, 
    "HTTPS Bad Chain": null, 
    "HTTPS Bad Hostname": null, 
    "HTTPS Expired Cert": null, 
    "Live": true, 
    "Redirect": false, 
    "Redirect To": null, 
    "Strictly Forces HTTPS": false, 
    "Valid HTTPS": true, 
    "endpoints": {
      "http": {
        "headers": {
          "Connection": "close", 
[...]

PR #9, opened 2016-08-17
Add redirect fields, canonical endpoint field, and rework HTTP probing logic

konklone:
This does a few things:

* Adds some fields describing the nature of the immediate and eventual redirects: whether the redirects go to an `https://` URI, whether it goes to a `www` subdomain, whether the redirect leaves the hostname, and whether the redirect leaves the hostname's parent (registerable) domain.
* Adds a `Canonical URL` field to the CSV and JSON output, that is determined using a port of [the logic from site-inspector](https://github.com/benbalter/site-inspector/blob/erics-mode/lib/site-inspector.rb#L201-L305). This makes use of the extra redirect fields mentioned above.
* Changes the way that HTTP/HTTPS endpoints are inspected, by first hitting each endpoint with following redirects disabled. If a redirect is detected, the endpoint is pinged again with following redirects enabled, and then information about the eventual redirect is parsed. This creates some additional requests compared to the previous method, but greatly simplifies the logic, and allows the exception handling to be limited more tightly.
* If an HTTPS endpoint throws an `SSLError` upon hitting it, it's retried again with certificate verification turned off. If it then works, the TLS error is overlooked and the endpoint's behavior is processed and analyzed. If it still doesn't work, then it's assumed to be a protocol error or some other novel issue that is preventing the connection from being established, and the endpoint is labeled not live.
* When redirects are followed, they are followed with certificate validation turned off. This greatly reduces the risk of having a certificate error get thrown when following redirects starting from an `http:` endpoint. Detecting a certificate error there during that process isn't of any real value to us, as all we care about is the eventual redirect destination URI.
* This refactoring to HTTP probing logic also removes the generic `except:` that was swallowing all errors (even `Ctrl-C` interrupts).
* Unfortunately, I had to **disable requests caching** (the `--cache` option now does nothing) because of a bug in how `requests-cache` handles redirects and the `follow_redirects` and `verify` parameters. I've filed this bug here: https://github.com/reclosedev/requests-cache/issues/70 To re-enable caching, we'll either need to fix the bug in requests-cache, or route around it somehow by using another method of writing our own caching layer.

### DIFF
--- a/pshtt.py
+++ b/pshtt.py
@@ -40,7 +40,7 @@ HEADERS = [
     "Domain", "Live", "Redirect",
     "Valid HTTPS", "Defaults HTTPS", "Downgrades HTTPS",
     "Strictly Forces HTTPS", "HTTPS Bad Chain", "HTTPS Bad Host Name",
-    "Expired Cert", "Weak Signature Chain", "HSTS", "HSTS Header",
+    "Expired Cert", "HSTS", "HSTS Header",
     "HSTS Max Age", "HSTS All Subdomains", "HSTS Preload",
     "HSTS Preload Ready", "HSTS Preloaded"
 ]
@@ -169,13 +169,12 @@ def basic_check(endpoint):
         location_header = req.headers.get('Location')
         # Absolute redirects (e.g. "https://example.com/Index.aspx")
         if location_header.startswith("http:") or location_header.startswith("https:"):
-            endpoint.redirect_immediately_to = location_header
+            immediate = location_header
 
         # Relative redirects (e.g. "Location: /Index.aspx").
         # Construct absolute URI, relative to original request.
         else:
-            endpoint.redirect_immediately_to = urlparse.urljoin(endpoint.url, location_header)
-
+            immediate = urlparse.urljoin(endpoint.url, location_header)
 
         # Chase down the ultimate destination, ignoring any certificate warnings.
         # TODO: try/except block on this request, and have checks expect possible None.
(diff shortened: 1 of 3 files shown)


pshtt.py
```diff
@@ -10,13 +11,19 @@
 import os
 import utils
 import logging
-import requests_cache
 
-from models import Domain, Endpoint
+try:
+    from urllib import parse as urlparse # Python 3
+except ImportError:
+    import urlparse # Python 2
```
konklone: Even though this tool currently requires Python 2, I made the new `urlparse` import agnostic so that it wouldn't be a barrier if this tool can eliminate its Python 2 dependency.

pshtt.py
```diff
@@ -30,10 +37,10 @@
 # The fields we're collecting, will be keys in JSON and
 # column headers in CSV.
 HEADERS = [
-    "Domain", "Live", "Redirect",
+    "Domain", "Canonical URL", "Live", "Redirect",
     "Valid HTTPS", "Defaults HTTPS", "Downgrades HTTPS",
     "Strictly Forces HTTPS", "HTTPS Bad Chain", "HTTPS Bad Host Name",
-    "Expired Cert", "Weak Signature Chain", "HSTS", "HSTS Header",
+    "Expired Cert", "HSTS", "HSTS Header",
```
konklone: I removed the `Weak Signature Chain` field, as I'd removed the code that produced it in a prior PR. Along with the `Canonical URL` field, this will mean some updated processing on DHS' side.

pshtt.py
```diff
@@ -101,87 +102,125 @@ def result_for(domain):
...
+
+        location_header = req.headers.get('Location')
+        # Absolute redirects (e.g. "https://example.com/Index.aspx")
+        if location_header.startswith("http:") or location_header.startswith("https:"):
+            immediate = location_header
+
+        # Relative redirects (e.g. "Location: /Index.aspx").
+        # Construct absolute URI, relative to original request.
+        else:
+            immediate = urlparse.urljoin(endpoint.url, location_header)
```
konklone: Sometimes you get redirects to relative locations (e.g. `Location: /Index.aspx`), and so if you're analyzing the URI then you need to join it to the original hostname the redirect is from.

pshtt.py
```diff
@@ -101,87 +102,125 @@ def result_for(domain):
...
+
+        # Chase down the ultimate destination, ignoring any certificate warnings.
+        # TODO: try/except block on this request, and have checks expect possible None.
+        ultimate_req = ping(endpoint.url, allow_redirects=True, verify=False)
```
konklone: This probably needs a `try`/`except` block around it too to handle connection errors when following redirects, even though we know that the endpoint is live at this point.

pshtt.py
```diff
@@ -209,32 +248,155 @@ def hsts_check(endpoint):
...
+            ):
+            endpoint.https_bad_chain = True
+
+        # Check for whether the hostname validates.
+        if (
+            (("Hostname Validation") in msg) and
+            (("FAILED") in msg) and
+            (("Certificate does NOT match") in msg)
+            ):
+            endpoint.https_bad_hostname = True
```
konklone: [I also improved the sslyze parsing logic to:

* Make the text parsing more strict in detecting each certificate error situation. Previously, the `CA Store` case would match both an invalid certifciate chain, and an expired certificate.
* Make it possible for multiple errors to be flagged. Previously, it was a set of conditionals (`if`/`elif`) that would only match the first error seen. This will now detect if a certificate has, for example, a bad chain *and* a bad name. (For context, this does happen -- we had [a bad bug in Pulse](https://github.com/18F/pulse/pull/224) where we weren't properly detecting this case, and fixing it forced us to drop the `Uses HTTPS` column by **4.5%**.

pshtt.py
```diff
@@ -270,26 +432,26 @@ def supports_https(endpoint):
...
-        return (http.redirect and (http.redirect_to[:5] == "https")) or (httpwww.redirect and (httpwww.redirect_to[:5] == "https"))
+        return (http.redirect and (http.redirect_eventually_to[:5] == "https")) or (httpwww.redirect and (httpwww.redirect_eventually_to[:5] == "https"))
     else:
         return False
 
 
 # Domain downgrades if https endpoint redirects to http
 def is_downgrades_https(http, httpwww, https, httpswww):
-    return (https.redirect and (https.redirect_to[:5] == "http:")) or (httpswww.redirect and (httpswww.redirect_to[:5] == "http:"))
+    return (https.redirect and (https.redirect_eventually_to[:5] == "http:")) or (httpswww.redirect and (httpswww.redirect_eventually_to[:5] == "http:"))
```
konklone: I generally renamed the field `redirect_to` to `redirect_eventually_to`, to be explicit.

pshtt.py
```diff
@@ -349,6 +511,12 @@ def is_hsts_preloaded(domain):
     return domain in preload_list
 
 
+# For "x.y.domain.gov", return "domain.gov".
+# TODO: use Public Suffix list to do this properly.
+def parent_domain_for(hostname):
+    return str.join(".", hostname.split(".")[-2:])
```
konklone: Noting here that this logic is naive, and that to properly detect the "registerable" parent domain of a hostname requires using the Public Suffix list. This is not necessary for `.gov`, but for the tool to be generically useful, it will be required. I described this in #8.

h-m-f-t:
:guitar: Most excellent!

PR #11, opened 2016-08-18
Comprehensive logic update

konklone:
This ports over a fair amount of logic from site-inspector for making domain-wide judgments, using the "canonical URL" logic added in #9.

A number of decisions use the "canonical HTTPS endpoint", which is to say that if the "canonical URL" for a domain is `http://www.agency.gov`, then the "canonical HTTPS endpoint" is `https://www.agency.gov`. This is meant to evaluate the state of HTTPS as it relates to the domain's "intended" URL.

* A domain has "Valid HTTPS" if its canonical HTTPS endpoint is live and uses a valid certificate.
* A domain defaults to HTTPS if its canonical URL is HTTPS.
* A domain downgrades HTTPS if the canonical HTTPS endpoint immediately redirects to **any** HTTP URI.
* A domain strictly forces HTTPS if both of its HTTP endpoints are either down or immediately redirect to **any** HTTPS URI.
* A domain has a bad chain, bad hostname, or expired certificate if its canonical HTTPS endpoint has a bad chain, bad hostname, or expired certificate.
* A domain has HSTS enabled if its canonical HTTPS endpoint has HSTS enabled. 
* The HSTS header and max age for the domain are sourced from the canonical HTTPS endpoint's HSTS header.
* A domain has HSTS enabled for the entire domain if its **root** HTTPS endpoint has HSTS enabled and uses `includeSubDomains`.
* A domain is HSTS "preload ready" if its **root** HTTPS endpoint has HSTS enabled, a max-age of at least 18 weeks, and uses `includeSubDomains` and the `preload` flag.
* A domain is HSTS preloaded if its domain name appears in the Chrome preload list, regardless of what header is present on any endpoint.

I also made a few changes to the CSV fields:

* I renamed `HSTS All Subdomains` to `HSTS Entire Domain`, to be explicit that this is measuring whether or not the HSTS policy includes the entire zone. This field will ignore HSTS headers set on `www` subdomains, regardless of whether that subdomain is canonical.
* I renamed `Expired Certificate` to `HTTPS Expired Certificate`, to be consistent with `HTTPS Bad Chain` and `HTTPS Bad Host Name`.
* I removed `HSTS Preload`, which was a domain-level assessment of whether the `preload` flag is set. It was not published or used in `site-inspector` or `domain-scan`, and can be confusing, because sometimes sites use a `preload` flag without an `includeSubDomains` field, and we're not interested in treating those any differently from an HSTS header with neither flag.

This PR brings `pshtt` to what I hope is near-parity with `site-inspector`. There are still some outstanding issues:

* The HSTS parser is too informal and doesn't handle pathological headers. 
* The HSTS parser also doesn't factor in `max-age=0`, which is the mechanism in HSTS to **disable** HSTS for a domain. While not used frequently, this is in use in the `.gov` space.
* There's not as much resiliency to pathological data in general. I'm in the process of running this version of the code on every federal `.gov`, and will note where it crashes or returns invalid data in some way. 
* Many fields will return `null` instead of `false`, and should have their values set explicitly by the main code path. IMO, this is preferable to giving fields default values at the model level, as it makes it difficult to detect bugs and places in the code path that aren't being exercised that should be. This is not an urgent issue, as `domain-scan` and DHS' renderer can understand this and treat `null` the same as `false`, but for general use this should be made consistent and the result should never include `null`'s.

There are some potential outstanding bugs in site-inspector/Pulse logic that I'd like to tackle as well, but they are minor and are more worth tackling once this tool is in use in both environments, rather than move the goalposts before we have parity.

### DIFF
--- a/models.py
+++ b/models.py
@@ -41,10 +41,12 @@ class Endpoint:
         self.redirect_immediately_to = None
         self.redirect_immediately_to_www = None
         self.redirect_immediately_to_https = None
+        self.redirect_immediately_to_http = None
         self.redirect_immediately_to_external = None
         self.redirect_immediately_to_subdomain = None
         self.redirect_eventually_to = None
         self.redirect_eventually_to_https = None
+        self.redirect_eventually_to_http = None
         self.redirect_eventually_to_external = None
         self.redirect_eventually_to_subdomain = None
 
@@ -82,9 +84,11 @@ class Endpoint:
             'redirect_immediately_to': self.redirect_immediately_to,
             'redirect_immediately_to_www': self.redirect_immediately_to_www,
             'redirect_immediately_to_https': self.redirect_immediately_to_https,
+            'redirect_immediately_to_http': self.redirect_immediately_to_http,
             'redirect_immediately_to_external': self.redirect_immediately_to_external,
             'redirect_immediately_to_subdomain': self.redirect_immediately_to_subdomain,
             'redirect_eventually_to_https': self.redirect_eventually_to_https,
+            'redirect_eventually_to_http': self.redirect_eventually_to_http,
             'redirect_eventually_to_external': self.redirect_eventually_to_external,
             'redirect_eventually_to_subdomain': self.redirect_eventually_to_subdomain
         }
--- a/models.py
+++ b/models.py
@@ -27,7 +27,7 @@ class Endpoint:
     def __init__(self, protocol, host, base_domain):
         # Basic endpoint description
         self.protocol = protocol
-        self.host = host # "www" or "root"
+        self.host = host  # "www" or "root"
         self.base_domain = base_domain
         self.url = self.url_for(protocol, host, base_domain)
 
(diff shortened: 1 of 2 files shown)


konklone:
Note this is not merge-ready yet. I need to:

- [ ] Fix the build (`flake8` errors).
- [ ] Evaluate why the scanner is failing on some domains (resilience issues, or logic bugs).

konklone:
OK, I've fixed the build (`flake8` warnings), and I've added a bunch of resiliency to HSTS parsing and connection failures. I ended up filing two bugs with underlying libraries as a result:

1. https://github.com/nabla-c0d3/sslyze/issues/167 <-- affects `fttesttwai.gov` and `qatesttwai.gov`
2. https://github.com/shazow/urllib3/issues/950 <-- affects `consumersentinel.gov`

I've hot-fixed (1) by catching `TypeError`s during `sslyze` operation, and will remove the hotfix if the issue is patched. I've left (2) open, because hot-fixing would be more challenging and because they've already indicated interest in fixing it.

This PR is ready for review and merge. The next step is take a CSV of scan results and compare them to Pulse's.

h-m-f-t:
:us:

Issue #12, opened 2016-08-18
Use more formal/complete HSTS parser


konklone:
Ideally, there would already be a good tested Python library for doing this. 

For reference, here's the parser I wrote in Ruby (with careful attention to the HSTS RFC):
https://github.com/benbalter/site-inspector/blob/erics-mode/lib/site-inspector.rb#L28-L84

konklone:
Some examples of pathological HSTS situations as I find them:

```
$ curl --head --header "User-agent: Hey there" https://telework.govHTTP/1.1 200 OK
Server: Microsoft-IIS/6.0
X-UA-Compatible: IE=edge
X-Powered-By: ASP.NET
Origin-Server: www1
Origin-Loc: m
Content-Type: text/html; charset=utf-8
Content-Length: 32104
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
x-frame-options: SAMEORIGIN
Strict-Transport-Security: max-age=31536000
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
x-frame-options: SAMEORIGIN
Strict-Transport-Security: max-age=31536000
Cache-Control: private, max-age=45689
Date: Thu, 18 Aug 2016 16:21:40 GMT
Connection: keep-alive
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
x-frame-options: SAMEORIGIN
Strict-Transport-Security: max-age=31536000
```

konklone:
```
$ curl --head --insecure https://jamesmadison.gov
HTTP/1.1 200 OK
Date: Thu, 18 Aug 2016 17:45:09 GMT
Server: Apache
Vary: Accept-Encoding
X-SA-Server: 1.apache.prod
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Strict-Transport-Security: includeSubDomains
Connection: close
Content-Type: text/html; charset=UTF-8
```

konklone:
Here are the test cases I wrote for `site-inspector` that I would want to reproduce in a unit test suite here:

https://github.com/benbalter/site-inspector/blob/erics-mode/test/test_hsts.rb

konklone:
```
$ curl --head https://www.bnl.gov
HTTP/1.1 302 Moved Temporarily
Cache-Control: Private
...
Strict-Transport-Security: 'max-age=10886400'
```

konklone:
```
[...]

PR #16, opened 2016-08-20
Bug fixing, obtain close parity with Pulse's existing scans

konklone:
This PR brings `pshtt` into "close enough" parity with the existing scans that take on [Pulse](https://pulse.cio.gov), which currently use [`site-inspector`](https://github.com/benbalter/site-inspector) v1.0.2.

I'll comment inline on the PR with the changes that were made here. It was a lot of little paper cuts to fix regressions and mistakes that became obvious when comparing the output of `pshtt` and `site-inspector` on ~1,300 `.gov` domains. I used `domain-scan` (on [this open PR branch](https://github.com/18F/domain-scan/pull/76) to orchestrate `pshtt` and `site-inspector` in tandem and then compared the delta between the two in a spreadsheet viewer.

One notable change:

* I added a `Redirect To` field, for parity with the current output of domain-scan's `inspect` scanner, which for "redirect domains" (domains where `Redirect` is true) contains the destination the domain redirects to.

I plan to take the version of `pshtt` on this branch and integrate it into Pulse imminently, replacing the current `site-inspector` path.

cc @garrettr @gbinal

### DIFF
--- a/pshtt.py
+++ b/pshtt.py
@@ -173,7 +173,7 @@ def basic_check(endpoint):
         return
 
     # Endpoint is live, analyze the response.
-    endpoint.headers = dict(req.headers)
+    endpoint.headers = req.headers
 
     endpoint.status = req.status_code
     if str(endpoint.status).startswith('3'):
--- a/pshtt.py
+++ b/pshtt.py
@@ -438,9 +438,6 @@ def canonical_endpoint(http, httpwww, https, httpswww):
     all_http_unused = http_unused(http) and http_unused(httpwww)
     at_least_one_http_upgrades = http_upgrades(http) or http_upgrades(httpwww)
 
-    print(http_upgrades(http))
-    print(http_upgrades(httpwww))
-
     is_https = (
         at_least_one_https_endpoint and
         all_http_unused and
--- a/pshtt.py
+++ b/pshtt.py
@@ -154,7 +154,7 @@ def basic_check(endpoint):
             return
         except requests.exceptions.RequestException:
             endpoint.live = False
-            logging.warn("Unexpected requests exception during retry.")
+            logging.warn("Unexpected requests exception during retry. Printing error:")
             logging.warn(utils.format_last_exception())
             return
 
@@ -307,6 +307,10 @@ def https_check(endpoint):
         logging.warn("sslyze exception parsing issuer, see https://github.com/nabla-c0d3/sslyze/issues/167")
         return
 
+    # Debugging
+    for msg in cert_response:
+        print(msg)
+
     # A certificate can have multiple issues.
     for msg in cert_response:
 
@@ -427,12 +431,15 @@ def canonical_endpoint(http, httpwww, https, httpswww):
     def http_upgrades(endpoint):
         return (
             endpoint.redirect_immediately_to_https and
-            (not http.redirect_immediately_to_external)
+            (not endpoint.redirect_immediately_to_external)
         )
 
     at_least_one_https_endpoint = https_used(https) or https_used(httpswww)
     all_http_unused = http_unused(http) and http_unused(httpwww)
     at_least_one_http_upgrades = http_upgrades(http) or http_upgrades(httpwww)
+
+    print(http_upgrades(http))
+    print(http_upgrades(httpwww))
 
     is_https = (
         at_least_one_https_endpoint and
--- a/pshtt.py
+++ b/pshtt.py
@@ -150,10 +150,12 @@ def basic_check(endpoint):
         except requests.exceptions.SSLError:
             # If it's a protocol error or other, it's not live.
             endpoint.live = False
+            logging.warn("Unexpected SSL protocol (or other) error during retry.")
             return
         except requests.exceptions.RequestException:
             endpoint.live = False
             logging.warn("Unexpected requests exception during retry.")
+            logging.warn(utils.format_last_exception())
             return
 
         # If it was a certificate error of any kind, it's live.
--- a/pshtt.py
+++ b/pshtt.py
@@ -371,38 +371,37 @@ def canonical_endpoint(http, httpwww, https, httpswww):
 
     at_least_one_www_used = httpswww.live or httpwww.live
 
+    def root_unused(endpoint):
+        return (
+            endpoint.redirect or
+            (not endpoint.live) or
+            endpoint.https_bad_hostname or  # harmless for http endpoints
+            (not str(endpoint.status).startswith("2"))
+        )
+
+    def root_down(endpoint):
+        return (
+            (not endpoint.live) or
+            endpoint.https_bad_hostname or
+            (not str(endpoint.status).startswith("2"))
+        )
+
+    def goes_to_www(endpoint):
+        return (
+            endpoint.redirect_immediately_to_www and
+            (not endpoint.redirect_immediately_to_external)
+        )
+
+    all_roots_unused = root_unused(https) and root_unused(http)
+
+    all_roots_down = root_down(https) and root_down(http)
+
     is_www = (
-        at_least_one_www_used and (
-            (
-                https.redirect or
-                (not https.live) or
-                https.https_bad_hostname or
-                (not str(https.status).startswith("2"))
-            ) and (
-                http.redirect or
-                (not http.live) or
-                (not str(http.status).startswith("2"))
-            )
-        ) and (
-            (
-                (
-                    (not https.live) or
-                    https.https_bad_hostname or
-                    (not str(https.status).startswith("2"))
-                ) and
-                (
-                    (not http.live) or
-                    (not str(http.status).startswith("2"))
-                )
-            ) or
-            (
-                https.redirect_immediately_to_www and
-                (not https.redirect_immediately_to_external)
-            ) or
-            (
-                http.redirect_immediately_to_www and
-                (not http.redirect_immediately_to_external)
-            )
+        at_least_one_www_used and
+        all_roots_unused and (
+            all_roots_down or
+            goes_to_www(https) or
+            goes_to_www(http)
         )
     )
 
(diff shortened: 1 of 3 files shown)


pshtt.py
```diff
@@ -37,8 +37,8 @@
 # The fields we're collecting, will be keys in JSON and
 # column headers in CSV.
 HEADERS = [
-    "Domain", "Canonical URL", "Live", "Redirect",
-    "Valid HTTPS", "Defaults HTTPS", "Downgrades HTTPS", "Strictly Forces HTTPS",
+    "Domain", "Canonical URL", "Live", "Redirect", "Redirect To",
+    "Valid HTTPS", "Defaults to HTTPS", "Downgrades HTTPS", "Strictly Forces HTTPS",
```
konklone: I added a `Redirect To` field as explained in the description, and also renamed a field to `Defaults to HTTPS` to make it the same as the field name in previous CSVs.

pshtt.py
```diff
@@ -116,7 +117,7 @@ def ping(url, allow_redirects=False, verify=True):
         verify=verify,
 
         # set by --user_agent
-        data={'User-Agent': USER_AGENT},
+        headers={'User-Agent': USER_AGENT},
```
konklone: This fixes a bug where the user agent was not actually being sent. The `data` parameter submits form data, not custom headers. This was causing us to submit the `User-Agent` string as a form data submission, which also added headers to the request and provided data that's not usually present on a `GET` request. While most servers ignored it, this caused Google's front-end servers to drop the traffic and return an error (instead of a redirect), which is how I noticed this issue.

pshtt.py
```diff
@@ -149,13 +150,17 @@ def basic_check(endpoint):
         except requests.exceptions.SSLError:
             # If it's a protocol error or other, it's not live.
             endpoint.live = False
+            logging.warn("Unexpected SSL protocol (or other) error during retry.")
             return
         except requests.exceptions.RequestException:
             endpoint.live = False
-            logging.warn("Unexpected requests exception during retry.")
+            logging.warn("Unexpected requests exception during retry. Printing error:")
+            logging.warn(utils.format_last_exception())
```
konklone: Now displaying generic TLS connection errors to console during the run, even without `--debug` or `-d`.

pshtt.py
```diff
@@ -149,13 +150,17 @@ def basic_check(endpoint):
...
             return
 
         # If it was a certificate error of any kind, it's live.
+        endpoint.live = True
+
```
konklone: Amid all the try/except blocks and conditionals, I wasn't proactively marking `endpoint.live` as `True` in one of them, which occasionally caused an endpoint to get marked as being down and for the canonical endpoint to change.

pshtt.py
```diff
@@ -262,6 +267,10 @@ def hsts_check(endpoint):
     if "max-age" in header.lower():
         endpoint.hsts_max_age = int(temp[0][len("max-age="):])
 
+    if endpoint.hsts_max_age <= 0:
+        endpoint.hsts = False
+        return
```
konklone: If an HSTS `max-age` is 0, this is the mechanism in the HSTS spec for disabling HSTS. Clients are supposed to clear their HSTS status for a site if they see `max-age=0`. I don't know offhand what a negative `max-age` means, but it would be nonsensical enough that I'm comfortable also saying that HSTS isn't present if that's the case.

pshtt.py
```diff
@@ -310,10 +323,11 @@ def https_check(endpoint):
             endpoint.https_expired_cert = True
 
         # Check for whether there's a valid chain to Mozilla.
+        # Note: this will also catch expired certs, but this is okay.
         if (
             (("Mozilla NSS CA Store") in msg) and
             (("FAILED") in msg) and
-            (("unable to get local issuer certificate") in msg)
+            (("Certificate is NOT Trusted") in msg)
```
konklone: This was previously not detecting the use of a self-signed certificate. I've made the catching logic more general for chain errors, at the cost of also including expired certificates. So if a certificate is expired, we'll now flag both the `HTTPS Bad Chain` _and_ `HTTPS Expired Cert` fields as `False`. I'm fine with that, personally.

pshtt.py
```diff
@@ -402,29 +418,30 @@ def canonical_endpoint(http, httpwww, https, httpswww):
     # It allows a site to be canonically HTTPS if the cert has
     # a valid hostname but invalid chain issues.
 
-    is_https = (
-        (
-            (https.live and (not https.https_bad_hostname)) or
-            (httpswww.live and (not https.https_bad_hostname))
```
konklone: I refactored this whole section on detecting whether the domain is canonically HTTPS, but you can see in the line above that there was a bug that prompted this, where I was checking `https.https_bad_hostname` for both `https` and `httpswww`, which was a bug and was causing issues. Refactoring this enabled me to more easily track down and fix this inconsistency.

pshtt.py
```diff
@@ -402,29 +418,30 @@ def canonical_endpoint(http, httpwww, https, httpswww):
...
+        )
+
+    at_least_one_https_endpoint = https_used(https) or https_used(httpswww)
+    all_http_unused = http_unused(http) and http_unused(httpwww)
+    at_least_one_http_upgrades = http_upgrades(http) or http_upgrades(httpwww)
+
+    is_https = (
+        at_least_one_https_endpoint and
+        all_http_unused and
+        at_least_one_http_upgrades
```
konklone: This refactored approach to calculating whether a domain's canonical endpoint uses HTTPS should be identical to the intended logic before, but should be easier to read, understand, and debug.

pshtt.py
```diff
@@ -481,6 +499,16 @@ def is_redirect(domain):
...
 
 
+# If a domain is a "redirect domain", where does it redirect to?
+def redirects_to(domain):
+    canonical = domain.canonical
+
+    if is_redirect(domain):
+        return canonical.redirect_eventually_to
+    else:
+        return None
```
konklone: The `Redirect To` field is populated by whatever eventual redirect destination the domain's canonical endpoint uses. This is meant to surface what a "redirect domain" is intended to redirect users to.

pshtt.py
```diff
@@ -520,7 +548,11 @@ def is_downgrades_https(domain):
     else:
         canonical_https = https
 
-    return (supports_https and canonical_https.redirect_immediately_to_http)
+    return (
+        supports_https and
+        canonical_https.redirect_immediately_to_http and
+        (not canonical_https.redirect_immediately_to_external)
+    )
```
konklone: Our logic was differing from `site-inspector` here in a way I failed to notice at first -- we only want to ding a domain for downgrading HTTPS if it downgrades *internally*. If some redirect domain (e.g. `https://redirect.gov`) is doing all it can to enforce HTTPS for itself, but has to redirect to some other domain that doesn't have HTTPS set up yet, that's the fault of the domain it's redirecting to, and will be surfaced in that domain or subdomain's report instead of the redirector's.

So this fix above modifies the downgrade detection logic to only care if the canonical endpoint downgrades to HTTP within the domain.

pshtt.py
```diff
@@ -628,7 +660,7 @@ def is_hsts_entire_domain(domain):
 def is_hsts_preload_ready(domain):
     https = domain.https
 
-    eighteen_weeks = (https.hsts_max_age and (https.hsts_max_age >= 10886400))
+    eighteen_weeks = ((https.hsts_max_age is not None) and (https.hsts_max_age >= 10886400))
```
konklone: This fixes a subtle bug where the value of `max-age` was getting returned as the value of `HSTS Preload Ready`, due to how Python logic short-circuiting works.

utils.py
```diff
@@ -4,6 +4,14 @@
...
 import logging
 import datetime
+import sys
+import traceback
+
+
+# Display exception without re-throwing it.
+def format_last_exception():
+    exc_type, exc_value, exc_traceback = sys.exc_info()
+    return "\n".join(traceback.format_exception(exc_type, exc_value, exc_traceback))
```
konklone: This is a generic function for outputting the value of an exception, in the same way that the system _would_ display the exception if it wasn't being caught. Helpful for seeing the traceback but not stopping execution.

konklone: I'll note that this function works in Python 2 and 3 both.

pshtt.py
```diff
@@ -256,12 +261,17 @@ def hsts_check(endpoint):
 
     # handle multiple HSTS headers, requests comma-separates them
     first_pass = re.split(',\s?', header)[0]
+    second_pass = re.sub('\'', '', first_pass)
```
konklone: I added a quick and dirty line to strip single quotes out of HSTS headers, which was causing a crash. I think these are actually not valid, raising the priority of tackling https://github.com/dhs-ncats/pshtt/issues/12#issuecomment-241179626.

pshtt.py
```diff
@@ -168,7 +173,7 @@ def basic_check(endpoint):
         return
 
     # Endpoint is live, analyze the response.
-    endpoint.headers = dict(req.headers)
+    endpoint.headers = req.headers
```
konklone: I'm now storing the original `requests.structures.CaseInsensitiveDict` form of the `headers` object on the `Endpoint` model rather than a converted `dict`. This is to make HSTS header extraction case-insensitive, at least under #12 is done and a proper header and HSTS parser is in place.

pshtt.py
konklone: I should note that the TLS connection errors get printed to `STDERR`, which mean that they don't get included in any tools that are using this tool's JSON output over `STDOUT` (such as `domain-scan`).

pshtt.py
```diff
@@ -349,38 +369,39 @@ def canonical_endpoint(http, httpwww, https, httpswww):
...
-            (
-                http.redirect_immediately_to_www and
-                (not http.redirect_immediately_to_external)
-            )
+        at_least_one_www_used and
+        all_roots_unused and (
+            all_roots_down or
+            goes_to_www(https) or
+            goes_to_www(http)
         )
```
konklone: The above refactoring of the `is_www` block should be identical to the version before, just easier to read and ebug.

h-m-f-t:
Clean, clear, and under control.

Thanks!

PR #22, opened 2016-08-21
Use "Hostname" instead of "Host Name"

konklone:
This is the correct way to spell it. Easiest to fix this now before it gets integrated into things.


h-m-f-t:
👍, I had changed the README but forgot to change it in pshtt.py. Thanks.

PR #36, opened 2016-10-14
Add 'Parent Domain', make most empty fields explicitly False

### LINKED ISSUE #8: Import public suffix list

h-m-f-t:
Though this doesn't resolve #8, this PR gives the tool a way to distinguish between a subdomain and a second-level domain. It also makes fields that return empty from a scan explicitly False (with 'Redirect To', 'HSTS Header', and 'HSTS Max Age' excepted, since an empty response is meaningful in those contexts).

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -37,7 +37,7 @@ TIMEOUT = 1
 # The fields we're collecting, will be keys in JSON and
 # column headers in CSV.
 HEADERS = [
-    "Domain", "Parent Domain", "Canonical URL", "Live", "Redirect", "Redirect To",
+    "Domain", "Base Domain", "Canonical URL", "Live", "Redirect", "Redirect To",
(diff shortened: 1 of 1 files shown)


konklone:
I know the method says `parent_domain_for`, but `Base Domain` is probably a bit more accurate, since `Parent Domain` might imply that for `x.y.z.com`, the parent is `y.z.com`, but we actually mean `z.com`. `Base Domain` would also be compatible with [domain-scan](https://github.com/18f/domain-scan)'s output. 

But I'm just nitpicking. :+1: either way.

h-m-f-t:
All your base, etc: `Base Domain` returns. Not strongly opinionated here, though the method name was what pshtt me over the edge.

The `parent_domain_for` method lives on.

PR #37, opened 2016-10-26
Check if any CSVs exists before copy

siccovansas:
If there are no CSVs present in the directory which is mounted to `/data` and if you do not provide and input to `pshtt` (e.g., if you only run `pshtt` or `pshtt -h` to find out about its functionality), then you will get the following error:
```
cp: cannot stat ‘/usr/src/app/*.csv’: No such file or directory
```
This change makes sure that the `cp` command is only executed if any CSVs are present.


h-m-f-t:
Looks good, thanks!

PR #38, opened 2016-10-27
Change UID and GID to USER_ID and GROUP_ID

siccovansas:
According to https://github.com/dhs-ncats/pshtt/blob/1d3db29e69b5af5d8f6dfdb40a2ab4bbc9cac1c7/entrypoint.sh the environment variables should be USER_ID and GROUP_ID instead of UID and GID, so this change updates README.md with the correct instructions.


h-m-f-t:
👍 👍   yup!

PR #41, opened 2016-11-21
Mark immediate external redirects as eventual redirects, when the eventual redirect fails

konklone:
This is meant to address situations like https://github.com/18F/pulse/issues/596 that come from scans like this:

https://s3.amazonaws.com/pulse.cio.gov/archive/2016-11-11/cache/pshtt/fedrooms.gov.json

Where each endpoint says `redirect` is `true`, but the overall domain comes up with a `Redirect` of `false`.

This happens when fedrooms.gov redirects to a gsa.gov URI that then redirects to another gsa.gov URI that is super flaky. When the final redirect doesn't work, fedrooms.gov's `redirect_eventually_*` fields are left as `null`. 

The overall `Redirect` field, which is trying to measure whether the domain is holistically a "redirect domain", correctly depends on whether the site eventually redirects away to an external domain.

To better capture information when later redirects don't work, and to avoid "punishing" a redirect domain that successfully redirects out to a flakey website, this will copy the immediate redirect information into the eventual redirect information for a domain.

Thoughts? Reasonable decision?

### DIFF
--- a/pshtt/cli.py
+++ b/pshtt/cli.py
@@ -68,5 +68,6 @@ def main():
         pshtt.csv_for(results, out_file)
         logging.warn("Wrote results to %s." % out_file)
 
+        
 if __name__ == '__main__':
     main()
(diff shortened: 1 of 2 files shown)


h-m-f-t:
I think this is a sensible fix. America thanks you, and I do, too.

konklone:
Awesome, thank you! :us:

Issue #29, opened 2016-09-07
Move some policy decisions into pshtt


konklone:
Right now, pulse.cio.gov makes some additional policy decisions on top of `pshtt` data to prepare its published values:
- Uses HTTPS: https://github.com/18F/pulse/blob/master/data/processing.py#L367-L381
- Enforces HTTPS: https://github.com/18F/pulse/blob/master/data/processing.py#L387-L420
- HSTS: https://github.com/18F/pulse/blob/master/data/processing.py#L431-L447
- Preloaded: https://github.com/18F/pulse/blob/master/data/processing.py#L451-L458

This logic should be moved down into `pshtt` so that others don't have to port the above logic to their own demesnes. And users who want to make different policy decisions will still have the direct measurements already in `pshtt` to do that with.

h-m-f-t:
@konklone: I'm sure there's a more sophisticated way to implement this than [this](https://github.com/dhs-ncats/pshtt/blob/fifteen-thirteenize/pshtt/pshtt.py#L105-L135) and would appreciate your take.

- `== False` is more readable than `!= True`, but e.g., `result['Valid HTTPS']` or `result['Downgrades HTTPS']` returns `None` for some domains that should otherwise be False, and it's not clear to me why or under what circumstances this occurs. (FWIW, I've been testing of late with a set of HHS domains with assorted compliance postures: cancer.gov, cdc.gov, clinicaltrials.gov, cms.gov, nlm.gov, psc.gov, surgeongeneral.gov.)
- It makes sense to me to have these fields not be on by default, only included by a `-m` or `--m1513` option. 
- Related to the first bullet, note the [failing build](https://travis-ci.org/dhs-ncats/pshtt/builds/176529046).

konklone:
> == False is more readable than != True, but e.g., result['Valid HTTPS'] or result['Downgrades HTTPS'] returns None for some domains that should otherwise be False, and it's not clear to me why or under what circumstances this occurs.

You should be more explicit with parentheses to ensure you get a predictable order of operations. Right now it's:

```
(
            is_downgrades_https(domain) != True and
            is_bad_chain(domain) and
            is_bad_hostname(domain) != True
)
```

And I think you should make it:

```
(
            (is_downgrades_https(domain) != True) and
            is_bad_chain(domain) and
            (is_bad_hostname(domain) != True)
)
```

[...]

PR #42, opened 2016-11-21
Move policy decisions into pshtt

h-m-f-t:
This PR fixes Issue #29, adding three new fields to the tool output: 

- **Domain Supports HTTPS**: A domain 'Supports HTTPS' when it doesn't downgrade and has valid HTTPS, or when it doesn't downgrade and has a bad chain but not a bad hostname. This is an attempt to capture whether the domain owner intentionally supports HTTPS: a bad chain shows an intent (though a misconfiguration), while a bad hostname often means support is not intended.
- **Domain Enforces HTTPS**: A domain that 'Enforces HTTPS' must 'Support HTTPS' and default to HTTPS. 'Redirect domains' must also strictly enforce HTTPS.
- **Domain Uses Strong HSTS**: A domain 'Uses Strong HSTS' when max-age ≥ 31536000.

These definitions closely follow the contours of M-15-13 compliance, and also match what [Pulse](https://pulse.cio.gov/https/agencies/#q=homeland%20security) calls "Uses HTTPS", "Enforces HTTPS", and "Strict Transport Security (HSTS)".

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -42,7 +42,7 @@ HEADERS = [
     "HTTPS Bad Chain", "HTTPS Bad Hostname", "HTTPS Expired Cert",
     "HSTS", "HSTS Header", "HSTS Max Age", "HSTS Entire Domain",
     "HSTS Preload Ready", "HSTS Preloaded",
-    "Uses HTTPS (M-15-13)", "Enforces HTTPS (M-15-13)", "HSTS (M-15-13)"
+    "Domain Supports HTTPS", "Domain Enforces HTTPS", "Domain Uses Strong HSTS"
 ]
 
 PRELOAD_CACHE = None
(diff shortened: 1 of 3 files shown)


pshtt/pshtt.py
```diff
@@ -99,7 +100,41 @@ def result_for(domain):
...
+        # or when it doesn't downgrade and has a bad chain but not a bad hostname.
+        # Domains with a bad chain "support" HTTPS but user-side errors should be expected.
+        'Domain Supports HTTPS': (
+            (is_downgrades_https(domain) != True) and
+            is_valid_https(domain)
+        ) or (
+            (is_downgrades_https(domain) != True) and
+            is_bad_chain(domain) and
+            (is_bad_hostname(domain) != True)
+        ),
```
konklone: I recommend (a) moving this boolean code to its own method (like the other methods are), and (b) reusing method calls to `is_defaults_to_https` and etc as needed, rather than copying the same logic here. This will make the code more maintainable and consistent.

konklone:
I added some changes in https://github.com/dhs-ncats/pshtt/pull/42/files/bb3461dd61a4388c581898f5a4f82e8f9dafc4c0..c9fc2e3acbbb3095241cc4f85121d502673b2fb7 that fix the `flake8` errors. 

It was mostly indentation, but I also changed some `[condition] != True` lines to `not [condition]`. This is consistent with the other conditional statements in the tool, and meets `flake8`. @h-m-f-t I didn't test out those changes, but could you 👁 and test them the same as you've been testing this code generally before merging?

h-m-f-t:
Looks good!

PR #43, opened 2016-12-06
Check whether a domain is formally pending HSTS preload (in queue)

konklone:
This adds a field, `HSTS Preload Pending`, which checks whether a domain is in the [HSTS preload queue](https://hstspreload.appspot.com/api/v2/pending).

Unfortunately, this does mean another HTTPS request made every time pshtt is run. It does this once already for the preload list itself. This is big enough to be cached on disk, but only if opted into with a flag, which I don't think anyone does right now.

For tools which orchestrate parallel uses of `pshtt`, like `domain-scan`, it might be helpful to have the orchestrator manage the downloading to disk of the preload list and pending list at the start of a bulk scan, and pass in their locations as flags so that `pshtt` don't have to re-download it every time during a bulk scan, but also don't rely on a cached version from a previous bulk scan.

What do you think? Should I add a cache flag for the pending list, like we do for the preload list? Is this field even worth capturing here at all?


h-m-f-t:
I think this makes sense and is a nice way to show an organization that a domain is making progress. With our HTTPS reports, we've had multiple teams reach out to say they've submitted their domain for preloading, and this gives them some validation.

Merging. Thanks!

PR #45, opened 2016-12-19
Fixes issue where no HTTP was causing Enforces to come up False

konklone:
Perhaps this was a regression from site-inspector, but pshtt was flagging Enforces as False is both HTTP endpoints were shut off entirely, which was not intended.

This also updates the HSTS preload URL to its new `hstspreload.org` location, to save on an HTTP redirect. (I meant to make it a second PR, but accidentally committed it alongside - seems trivial enough not to bother separating unless I'm asked to.)


h-m-f-t:
Good find. Thanks @konklone for the fix, and thanks to the USDA team for pointing out the error. 🇺🇸

PR #50, opened 2017-01-28
Fix incorrect null results for "Downgrades HTTPS"

### LINKED ISSUE #49: "Downgrades HTTPS" is null for abcnews.go.com
When you access `https://abcnews.go.com` in a browser, it downgrades the connection to HTTP; however, it does not use an HTTP redirect:

```
$ curl -I https://abcnews.go.com
HTTP/1.1 200 OK
[..snip..]
```

Instead, it uses Javascript to redirect:

```html
<script>
        if (window.location.protocol == "https:" && window.parent.location.hostname.indexOf("outbrain") == -1) {
                var _sslurl = window.location.href.replace("https://", "http://");
                window.location.replace(_sslurl);
                window.location.href = _sslurl;
        }
</script>
```

pshtt's results are confusing. `pshtt --json abcnews.go.com` returns `"Valid HTTPS": true` but also returns `"Downgrades HTTPS": null`. I would expect that if `"Valid HTTPS" true`, then `"Downgrades HTTPS"` should be one of true or false, but not null. null only makes sense if the site doesn't have valid HTTPS, because then the question of whether or not it downgrades it is moot.

What do y'all think? Is this a bug?

Note that I am *not* saying that pshtt should (necessarily) detect downgrades done with Javascript. While it would be possible (and possibly cool) to implement that, you would need to implement browser automation to do it right, which is a significant expansion of the crawler's current design. However, I do think there seems to be a bug in this case, where the result should be true or false (not null).

garrettr:
This commit fixes 2 bugs with the logic for computing "Downgrades HTTPS":

1. If a site **does not** have HTTPS (`"Valid HTTPS": false"`), then
"Downgrades HTTPS" should be `null` because it doesn't make sense to ask if a
site downgrades HTTPS if it doesn't have HTTPS in the first place. Previously,
the code would return `false` in this case.

2. If a site **does** have HTTPS, there were some cases where the previous code
would incorrectly return `null` instead of `false`. Specifically, this would
happen if a site's domain name was also the canonical endpoint, in which case
the site does not redirect and all values related to redirects are None. Due to
Python's short-circuit evaluation of boolean expressions, this None value was
returned instead of the intended False from `is_downgrades_https`.

Fixes #49.

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -588,11 +588,19 @@ def is_downgrades_https(domain):
     else:
         canonical_https = https
 
-    return (
-        supports_https and
-        canonical_https.redirect_immediately_to_http and
-        (not canonical_https.redirect_immediately_to_external)
-    )
+    # If the site doesn't support HTTPS, it doesn't make sense to ask if it
+    # downgrades it or not, so return None instead of True or False.
+    downgrades_https = None
+    if supports_https:
+        # Explicitly convert to bool to avoid unintentionally returning None,
+        # which may happen if the site doesn't redirect.
+        downgrades_https = bool(
+            canonical_https.redirect_immediately_to_http and
+            (not canonical_https.redirect_immediately_to_external)
+        )
+
+    return downgrades_https
+
 
 
 # A domain "Strictly Forces HTTPS" if one of the HTTPS endpoints is
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -588,18 +588,13 @@ def is_downgrades_https(domain):
     else:
         canonical_https = https
 
-    # If the site doesn't support HTTPS, it doesn't make sense to ask if it
-    # downgrades it or not, so return None instead of True or False.
-    downgrades_https = None
-    if supports_https:
-        # Explicitly convert to bool to avoid unintentionally returning None,
-        # which may happen if the site doesn't redirect.
-        downgrades_https = bool(
-            canonical_https.redirect_immediately_to_http and
-            (not canonical_https.redirect_immediately_to_external)
-        )
-
-    return downgrades_https
+    # Explicitly convert to bool to avoid unintentionally returning None,
+    # which may happen if the site doesn't redirect.
+    return bool(
+        supports_https and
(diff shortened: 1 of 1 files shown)


konklone:
`2.` is great, and describes the right fix for #49. Thank you for tracking this down and filing the PR!

However, I think we should return `false` even if a site does not have HTTPS. While there's a valid semantic difference, I don't think it's reasonable to oblige all downstream tools to potentially handle semantic differences between a `null` and `false` response. Instead, I think we should basically expunge all `null` fields from output and treat any emergence of them as a bug. If we were doing that from the get go, then I think #49 wouldn't have happened at all.

Would you mind tweaking the PR to just fix the short-circuit bug? Also, the [build is failing on a couple `flake8` errors](https://travis-ci.org/dhs-ncats/pshtt/builds/196069686), so fixing those would also make it easier to merge.

garrettr:
> However, I think we should return false even if a site does not have HTTPS. While there's a valid semantic difference, I don't think it's reasonable to oblige all downstream tools to potentially handle semantic differences between a null and false response.

Ok! I disagree, and think the semantic difference is useful to downstream tools. As the maintainer of a downstream tool, and frequent debugger of pshtt, I think it is helpful to distinguish between "true", "false", and "n/a". But it is not my place to debate that! I will make the change you requested.

Will also fix the flake8 errors.

konklone:
> I will make the change you requested.

I appreciate it! This looks good.

> Ok! I disagree, and think the semantic difference is useful to downstream tools. As the maintainer of a downstream tool, and frequent debugger of pshtt, I think it is helpful to distinguish between "true", "false", and "n/a".

That's good feedback, and I could be wrong. But let me put it another way -- if we keep it to `true`/`false`, downstream tools that want to can still apply logic to distinguish the case (as in, if `Valid HTTPS` is false, display `N/A` for `Downgrade HTTPS`). But if we add `null` values in, all downstream tools are forced to care about N/A vs False. That said, I only know of 3 live downstream users -- Pulse, SecureTheNews, and DHS reports -- so that population is probably who we should be optimizing for.

PR #60, opened 2017-03-16
Add check for "Base Domain HSTS Preloaded"

h-m-f-t:
This PR adds another field to pshtt output, which is (a bit clunkily) called `Base Domain HSTS Preloaded`. 

This checks the preload list for a given domain's parent domain (e.g., `dhs.gov` for `workplace.dhs.gov`). This is subtly different from `HSTS Entire Domain`, which checks whether HSTS is set correctly to encompass the entire zone-- a prerequisite to actually being preloaded. 

In compliance checks (like ones I do for [M-15-13](https://https.cio.gov/guide/#options-for-hsts-compliance)), a given domain can be compliant if it uses + enforces HTTPS AND its HSTS max-age is at least a year, OR if the parent domain is preloaded. This change enables easy checking on a per (sub)domain basis, where previously a base domain's preload status was solely reflected on that domain's output.

For example, `privacy.whitehouse.gov` does not itself use an HSTS header, and would fall afoul of M-15-13 compliance at the individual domain level-- but `whitehouse.gov` [is preloaded](https://chromium.googlesource.com/chromium/src/net/+/master/http/transport_security_state_static.json#3351). This enables quick and appropriate scoring for `privacy.whitehouse.gov`.

(This does not resolve our issue of not using the public suffix list.)

I've also un-blockquoted parts of the README, which was formatting weirdly.

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -42,7 +42,8 @@ HEADERS = [
     "HTTPS Bad Chain", "HTTPS Bad Hostname", "HTTPS Expired Cert",
     "HSTS", "HSTS Header", "HSTS Max Age", "HSTS Entire Domain",
     "HSTS Preload Ready", "HSTS Preload Pending", "HSTS Preloaded",
-    "Domain Supports HTTPS", "Domain Enforces HTTPS", "Domain Uses Strong HSTS"
+    "Base Domain HSTS Preloaded", "Domain Supports HTTPS",
+    "Domain Enforces HTTPS", "Domain Uses Strong HSTS"
 ]
 
 PRELOAD_CACHE = None
@@ -103,6 +104,7 @@ def result_for(domain):
         'HSTS Preload Ready': is_hsts_preload_ready(domain),
         'HSTS Preload Pending': is_hsts_preload_pending(domain),
         'HSTS Preloaded': is_hsts_preloaded(domain),
+        'Base Domain HSTS Preloaded': is_parent_hsts_preloaded(domain),
 
         'Domain Supports HTTPS': is_domain_supports_https(domain),
         'Domain Enforces HTTPS': is_domain_enforces_https(domain),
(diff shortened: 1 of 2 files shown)


pshtt/pshtt.py
```diff
@@ -720,6 +722,11 @@ def is_hsts_preloaded(domain):
     return domain.domain in preload_list
 
 
+# Whether a domain's parent domain is in Chrome's HSTS preload list.
+def is_parent_hsts_preloaded(domain):
+    return parent_domain_for(domain.domain) in preload_list
```
konklone: The only nit I have is here - I would call `is_hsts_preloaded()` on the result of `parent_domain_for(domain.domain)`, instead of also saying `in preload_list`, since that's an implementation detail (keeping a global list of domains cached up-front). 

That way, if we ever change the implementation of how we manage looking up preload status, we only have to change it in one place.

konklone: @h-m-f-t Just a note on my outstanding nit. ;) Should be a very quick refactoring change, if you concur.

h-m-f-t:
```python
def is_parent_hsts_preloaded(domain):
   return is_hsts_preloaded(parent_domain_for(domain.domain))
```
yields the following
```
Traceback (most recent call last):
  File "./cli.py", line 73, in <module>
    main()
  File "./cli.py", line 54, in main
    results = pshtt.inspect_domains(domains, options)
  File "/Users/cameron.dixon/Dev/pshtt/pshtt/pshtt.py", line 889, in inspect_domains
    results.append(inspect(domain))
  File "/Users/cameron.dixon/Dev/pshtt/pshtt/pshtt.py", line 71, in inspect
    return result_for(domain)
  File "/Users/cameron.dixon/Dev/pshtt/pshtt/pshtt.py", line 107, in result_for
    'HSTS Base Domain Preloaded': is_parent_hsts_preloaded(domain),
  File "/Users/cameron.dixon/Dev/pshtt/pshtt/pshtt.py", line 727, in is_parent_hsts_preloaded
    return is_hsts_preloaded(parent_domain_for(domain.domain))
  File "/Users/cameron.dixon/Dev/pshtt/pshtt/pshtt.py", line 722, in is_hsts_preloaded
    return domain.domain in preload_list
AttributeError: 'str' object has no attribute 'domain'
```

h-m-f-t:
Thanks team 🇺🇸 👍

PR #61, opened 2017-03-20
Added slant quotes to bash file existence check

egyptiankarim:
I tend to put ISO 8601 dates into the filenames of various pshtt scans I run. The '-' characters escape the -f file existence check that occurs. Wrapping the search string in slant quotes resolves the issue. I have a more robust fix I'm working on that uses a find command, but this fixes, what was for me a vexing problem, relatively quickly.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -27,7 +27,7 @@ else
 fi
 
 # Copy the input if there is one in the mount point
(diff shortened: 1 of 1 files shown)


egyptiankarim:
I'm going to hold off on a PR for this until I get the more robust file check working.

PR #62, opened 2017-03-20
Exception handling that prints "err" objects

egyptiankarim:
Added print outs of all the exception objects to support some debugging I was doing. Should probably be done in other places to keep things consistent, but I just hit the areas where I was running into issues.

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -155,15 +155,14 @@ def basic_check(endpoint):
         # Retry with certificate validation disabled.
         try:
             req = ping(endpoint.url, verify=False)
-        except requests.exceptions.SSLError:
+        except requests.exceptions.SSLError as err:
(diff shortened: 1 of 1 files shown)


konklone:
What I would prefer is that the stack traces be displayed in debug mode only. So we'd continue to `log.warn()` what we currently do, but then do another line and `log.debug()` the stack trace.

This should already Just Work, as we already configure logging at the debug level if `--debug` is passed to the CLI:

* https://github.com/dhs-ncats/pshtt/blob/master/pshtt/utils.py#L70-L75
* https://github.com/dhs-ncats/pshtt/blob/master/pshtt/cli.py#L33

@egyptiankarim Mind changing your PR to do the stack traces in `log.warn()` statements? I'm a :+1: for merging if so.

egyptiankarim:
@konklone, that totally works. I'll work on a separate PR.

PR #63, opened 2017-03-20
Remove superfluous file checks.

egyptiankarim:
I previously submitted a pull request to fix an issue I was having with bash doing these file existence checks, but realized later why it was dumb to use backquotes in bash to try and capture unescaped characters ( #61 ). 

I gave it some (read, a tiny little bit of) thought and realized that the file existence checks here are actually kind of superfluous. The <kbd>cp</kbd> will fail elegantly if no files match the *.csv pattern, so there's nothing really gained from the check. 

Further, by removing the check we can get around weird issues with "strangely" named files (e.g., ISO 8601 dated files, and files with white space in their names). We also get around weird issues with how the -f check works in various versions of bash.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -26,16 +26,13 @@ else
 	gid=1000
 fi
 
-# Copy the input if there is one in the mount point
-if [ -f /data/*.csv ]; then
-  cp /data/*.csv /usr/src/app
-fi
+# Copy the inputs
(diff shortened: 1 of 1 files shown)


konklone:
:+1: from me.

h-m-f-t:
Thanks @egyptiankarim :us:

Issue #66, opened 2017-03-30
Logic issue with is_redirect method


egyptiankarim:
The way is_redirect is currently written, a domain will get flagged as a redirect if all endpoints are down or otherwise code 400-ing. For example, imagine domain.tld, which only has its https endpoint up and happens to be returning for a 404 for its index. That will return `true` for is_redirect, which is confusing (and happening a bunch where I am).

I'm curious if that whole logic block can be simplified to check and see if any of the end points are returning a code 300. That entire block starting on line 527 might reasonably be reduced to something like:

```
redirection_codes = range(300, 309)

return (https.status in redirection_codes) or (http.status in redirection_codes) or (httpswww.status in redirection_codes) or (httpwww.status in redirection_codes)
```

I'll submit a pull request and we can hash it out there.

konklone:
From #67:

> I think we should add a conditional to the calculation that ensures that at least one of the endpoints is an external redirect. Something like:

```python
# ...
  and (
    httpwww.redirect_eventually_to_external or 
    http.redirect_eventually_to_external or 
    httpswww.redirect_eventually_to_external or 
    https.redirect_eventually_to_external
  )
```

I closed #67, and am moving the proposed resolution here for tracking.

konklone:
Just noting for myself and others that this is still an issue, and affects at least one second-level domain in exim.gov:

https://s3-us-gov-west-1.amazonaws.com/cg-4adefb86-dadb-4ecf-be3e-f1c7b4f6d084/live/cache/pshtt/exim.gov.json

PR #67, opened 2017-03-30
Modified the is_redirect method to check for code 300 server responses.

egyptiankarim:
This is in reference to #66. I've attempted to simplify the check here by just looking at the server response as opposed to the previous battery, which could provide confusing feedback in certain cases.

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -533,12 +533,7 @@ def is_redirect(domain):
     # 308 Permanent Redirect (RFC 7538)
     redirect_codes = [300, 301, 302, 303, 304, 305, 306, 307, 308]
 
-    return (
-        (https.status in redirect_codes) or
-        (http.status in redirect_codes) or
-        (httpswww.status in redirect_codes) or
-        (httpwww.status in redirect_codes) or
-    )
+    return ((https.status in redirect_codes) or (http.status in redirect_codes) or (httpswww.status in redirect_codes) or (httpwww.status in redirect_codes))
 
 
 # If a domain is a "redirect domain", where does it redirect to?


konklone:
I get what you're trying to do here, but this fundamentally changes what's being detected. Currently, pshtt's `is_redirect` method is trying to determine whether a domain's only purpose is to redirect away to some other domain -- in other words, if **all** endpoints redirect **externally**. This PR would cause pshtt to flag a domain if **any** of its endpoints redirect **anywhere**, and that's not desired.

This isn't common in most of the internet, but is very common in the USG for things like health campaigns or other awareness-raising issues. The shortness of `.gov` seems to serially tempt the White House and other agencies to brand an initiative as "something.gov" even if they plan to host it at a subdomain (for example, [ethics.gov](https://ethics.gov), and they want the "something.gov" to redirect you there.

This detection is useful for a number of things -- for HTTPS-tracking, we use it to give slightly different logic to conferring "Enforces HTTPS" if the domain is a redirect domain (we let its `http://` endpoints immediately redirect to external `https://` URLs, rather than first having to redirecting to one of its own `https://` endpoints first).

The behavior you're flagging in #66, though, is definitely a bug:

> The way is_redirect is currently written, a domain will get flagged as a redirect if all endpoints are down or otherwise code 400-ing. For example, imagine domain.tld, which only has its https endpoint up and happens to be returning for a 404 for its index. That will return true for is_redirect, which is confusing (and happening a bunch where I am).

I think we should add a conditional to the calculation that ensures that at least one of the endpoints is an external redirect. Something like:

```python
# ...
  and (
    httpwww.redirect_eventually_to_external or 
    http.redirect_eventually_to_external or 
    httpswww.redirect_eventually_to_external or 
    https.redirect_eventually_to_external
  )
```

Thoughts?

konklone:
Moved resolution tracking to #66.

PR #70, opened 2017-05-15
Update sslyze to 1.1.0

konklone:
This updates `pshtt` to use `sslyze` 1.10: https://github.com/nabla-c0d3/sslyze/releases/tag/1.1.0

This version of sslyze supports Python 3, but this PR **does not update pshtt to Python 3 yet**. This is just to:

* change out the sslyze 0.X syntax to sslyze 1.X syntax (which is now considered stable as of 1.0)
* to remove handling for an outdated error that sslyze fixed 
* to add handling for a new error that sslyze has not yet fixed

It also updates the `requests` version to the latest version, to fix an issue I encountered between a separate dependency and that older version of `requests`.

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -16,7 +16,6 @@ try:
 except ImportError:
     import urlparse  # Python 2
 
-import nassl
 import sslyze
 import sslyze.synchronous_scanner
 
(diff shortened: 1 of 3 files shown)


h-m-f-t:
👍

PR #74, opened 2017-06-12
Needs review: Python 3 support

konklone:
This PR takes advantage of `sslyze`'s new Python 3 support, and builds on #70 to allow `pshtt` to be run in Python 3.

From my testing, `pshtt` should also continue to work in Python 2 as well. However, we could decide to abandon support for Python 2.

Some things I haven't tested, and probably need review:

* Running the tool **without** a `pip install` workflow. I've only run it after doing `pip install -e .` in the root of the dir, and then running `pshtt` directly.
* The Dockerfile. I haven't updated it at all for Python 3.
* More code paths than just `pshtt -d -j example.com`

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 `pshtt` was developed to _push_ organizations— especially large ones like the US Federal Government :us: — to adopt HTTPS across the enterprise. Federal .gov domains must comply with [M-15-13](https://https.cio.gov), a 2015 memorandum from the White House Office of Management and Budget that requires federal agencies to enforce HTTPS on their web sites and services by the end of 2016. Much has been done, and [still more yet to do](https://18f.gsa.gov/2017/01/04/tracking-the-us-governments-progress-on-moving-https/).
 
-`pshtt` is a collaboration between the [Department of Homeland Security's National Cybersecurity Assessments and Technical Services (NCATS) team](https://github.com/dhs-ncats) and [the General Service Administration's 18F team](https://18f.gsa.gov), with contributions from NASA and various non-governmental organizations.
+`pshtt` is a collaboration between the [Department of Homeland Security's National Cybersecurity Assessments and Technical Services (NCATS) team](https://github.com/dhs-ncats) and [the General Service Administration's 18F team](https://18f.gsa.gov), with [contributions from NASA and various non-governmental organizations](https://github.com/dhs-ncats/pshtt/graphs/contributors).
 
 ## Getting Started
 
--- a/README.md
+++ b/README.md
@@ -3,12 +3,38 @@
 
 `pshtt` was developed to _push_ organizations— especially large ones like the US Federal Government :us: — to adopt HTTPS across the enterprise. Federal .gov domains must comply with [M-15-13](https://https.cio.gov), a 2015 memorandum from the White House Office of Management and Budget that requires federal agencies to enforce HTTPS on their web sites and services by the end of 2016. Much has been done, and [still more yet to do](https://18f.gsa.gov/2017/01/04/tracking-the-us-governments-progress-on-moving-https/).
 
-`pshtt` is a collaboration between GSA's 18F and the DHS National Cybersecurity Assessments and Technical Services team.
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -46,6 +46,10 @@
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.6',
```
konklone: Have we ever tested on Python 2.6? If not, I think we should remove it from `setup.py`.

konklone:
@h-m-f-t @garrettr @kaplan89 @siccovansas @egyptiankarim This PR adds Python 3 support to `pshtt`. 

Would some of you mind testing out this PR in your own use cases? In particular, I haven't touched the Dockerfile at all.

ultiferrago:
> Running the tool without a pip install workflow. I've only run it after doing pip install -e . in the root of the dir, and then running pshtt directly.

This does not appear to work as is. If it is an intended use case, will need modification.

konklone:
@ultiferrago I just verified that this approach to running the module locally (without requiring a `pip install -e .`) works as expected:

```
python -m pshtt.cli -d -j example.com
```

I think it's likely to be necessary to use the [`-m [module-name]` option](https://docs.python.org/3/using/cmdline.html), since the overall module is currently set up to use `pshtt/cli.py` as an entry point.

I've added a couple commits that touch up the documentation to make this explicit.

README.md
```diff
@@ -3,14 +3,40 @@
 
 `pshtt` was developed to _push_ organizations— especially large ones like the US Federal Government :us: — to adopt HTTPS across the enterprise. Federal .gov domains must comply with [M-15-13](https://https.cio.gov), a 2015 memorandum from the White House Office of Management and Budget that requires federal agencies to enforce HTTPS on their web sites and services by the end of 2016. Much has been done, and [still more yet to do](https://18f.gsa.gov/2017/01/04/tracking-the-us-governments-progress-on-moving-https/).
 
-`pshtt` is a collaboration between GSA's 18F and the DHS National Cybersecurity Assessments and Technical Services team.
+`pshtt` is a collaboration between the [Department of Homeland Security's National Cybersecurity Assessments and Technical Services (NCATS) team](https://github.com/dhs-ncats) and [the General Service Administration's 18F team](https://18f.gsa.gov), with [contributions from NASA and various non-governmental organizations](https://github.com/dhs-ncats/pshtt/graphs/contributors).
```
konklone: @h-m-f-t @egyptiankarim This is a separate topic from the main PR, and I can separate it if need be, but while I was editing the README I updated the collaboration text to link to things, order NCATS first, and to mention NASA's contributions.

README.md
```diff
@@ -3,14 +3,40 @@
...
+```bash
+pip install -r requirements.txt
+```
+
+Then run it as a module via `python -m`:
+
+```bash
+python -m pshtt.cli example.com [options]
+```
+
```
konklone: @ultiferrago The lines above adjust and expand the documentation to make clear how the tool can be run locally without installing as a module.

konklone:
@h-m-f-t @ultiferrago @egyptiankarim If folks are good with this PR, I think it's ready to merge. With the caveat that I haven't tested the Docker-based workflow, which I'm not personally familiar with.

README.md
ultiferrago: Confirmed, you are correct. That fixed the issues with relative imports.

setup.py
egyptiankarim: @konklone agreed.

konklone:
> As currently set, pshtt will not run in a 2.7 container, because well that's the whole point of the PR 😄

Hmm, @egyptiankarim actually I did intend to retain Python 2.7 support for the tool. Why does it break in a 2.7 container?

konklone:
> I did test locally with a 3.5-onbuild tag and my Docker container worked just as I would expect.

@egyptiankarim I updated the `Dockerfile` in this PR to pull from the `3.5-onbuild` tag. Want to test it out from this branch?

egyptiankarim:
@konklone My mistake. I can't reproduce the issue I was having with 2.7 now. I'm getting clean runs in containers tagged with either version. Sorry for the false alarm.

h-m-f-t:
Finally had a chance to review and test. LGTM. 👍

konklone:
Woo! Python 3 support achieved. ^_^

PR #75, opened 2017-06-16
Refactored base domain hsts preload check to reduce code reuse.

ultiferrago:
Used is_hsts_preloaded() to prevent duplicate code. 

Could potentially refactor is_hsts_preloaded() to utilize the hostname as a parameter rather than a Domain object as that is the only information needed by check. This would prevent the need from creating a new Domain object in is_parent_hsts_preloaded().


konklone:
Looks good to me! Thank you.

I really wish we had a test suite to verify things like this more rigorously, but that's a separate issue. ;)

Issue #33, opened 2016-09-25
If given a www. subdomain, remove the www. before scanning


konklone:
In other words, running `pshtt` on `www.nasa.gov` should be the same as running it on `nasa.gov`. The tool already makes `www` a special case -- that's a fundamental assumption of the entire tool -- so this is an appropriate way to normalize input.

This is especially relevant when checking fourth-level subdomains, e.g. `www.something.example.com`, which are unfortunately common and come up frequently in subdomain data.

konklone:
Fixed by #76.

PR #76, opened 2017-06-19
Scan and remove prepended www/https/http for input domains.

ultiferrago:
Addresses #33.

Scans for http/www/https and removes them from the domain name.

### DIFF
--- a/pshtt/cli.py
+++ b/pshtt/cli.py
@@ -41,7 +41,7 @@ def main():
     if args['INPUT'][0].endswith(".csv"):
         domains = utils.load_domains(args['INPUT'][0])
     else:
-        domains = args['INPUT']
+        domains = utils.format_domains(args['INPUT'])
 
     # If the user wants to sort them, sort them in place.
     if args['--sorted']:
--- a/pshtt/cli.py
+++ b/pshtt/cli.py
@@ -41,7 +41,9 @@ def main():
     if args['INPUT'][0].endswith(".csv"):
         domains = utils.load_domains(args['INPUT'][0])
     else:
-        domains = utils.format_domains(args['INPUT'])
+        domains = args['INPUT']
+
+    domains = utils.format_domains(domains)
 
     # If the user wants to sort them, sort them in place.
     if args['--sorted']:
--- a/pshtt/utils.py
+++ b/pshtt/utils.py
@@ -81,7 +81,7 @@ def format_domains(domains):
     formatted_domains = []
 
     for domain in domains:
-        # Replace a single instance of http://, and www. if present.
-        formatted_domains.append(domain.replace("http://", "", 1).replace("www.", "", 1))
+        # Replace a single instance of http://, https://, and www. if present.
+        formatted_domains.append(domain.replace("https://", "").replace("http://", "", 1).replace("www.", "", 1))
 
     return formatted_domains
--- a/pshtt/utils.py
+++ b/pshtt/utils.py
@@ -63,7 +63,7 @@ def load_domains(domain_csv):
 
             row[0] = row[0].lower()
 
-            domains.append(row[0])
+            domains.append(format_domain(row[0]))
     return domains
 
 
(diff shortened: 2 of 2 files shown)


pshtt/cli.py
```diff
@@ -41,7 +41,7 @@ def main():
     if args['INPUT'][0].endswith(".csv"):
         domains = utils.load_domains(args['INPUT'][0])
     else:
-        domains = args['INPUT']
+        domains = utils.format_domains(args['INPUT'])
```
konklone: I think it would be cleaner to make one call to `utils.format_domains()`, outside of this conditional, rather than calling it in one way for directly input domains and calling the `utils.format_domain()` function directly for domains loaded in via CSV.

This would also allow it to be one `utils.format_domains()` function, rather than two very similarly named functions.

pshtt/utils.py
```diff
@@ -75,3 +75,28 @@ def configure_logging(debug=False):
...
+    # Check for www. prepended
+    if domain_name.startswith("www."):
+        return domain_name[4:]
+    # Check for https prepended, account for www. as well
+    if domain_name.startswith("https://"):
+        return format_domain(domain_name[8:])
+
+    # Check for http:// prepended, account for www. as well
+    if domain_name.startswith("http://"):
+        return format_domain(domain_name[7:])
```
konklone: Instead of doing this recursively, I think it would be more predictable and safe to do this iteratively. I also think it would be more readable and clear to couple the detection of the prefix with the modification to the string itself, rather than truncate it using character indexes.

So first it should replace a regex of `^https?://` with an empty string, and then it should replace any remaining prefix of `^www\.` with an empty string. If we also merged this into `utils.format_domains()`, you would then append the result to the ongoing array and then return the whole thing once they're all done.

With the current recursive solution, `http://http://domain.com` would be turned into `domain.com` -- and despite `http://http://domain.com` being a broken domain, that is actually not what what we are trying to fix with this check. We're trying to offer a straightforward convenience to remove a `http://` or `https://` protocol prefix, and a single `www.` subdomain prefix if it exists.

ultiferrago: I am curious why we wouldn't want to correct to just the domain names without the prefixes? My interpretation of the fix was to remove any prefixed www or protocol. Is there a functional reason for only removing a single instance of this?

konklone: `http://domain.com` is a valid URL with a valid hostname contained inside it which we want to scan. `http://http://domain.com` isn't a valid URL or hostname, and it's not necessarily the case that the user's intent is to scan `domain.com`. 

In other words, we don't want to get into the business of silently "fixing" invalid data where we don't have information on what the intent of that data was. Instead, we should gracefully surface invalid URLs for correction by the upstream user.

konklone:
@ultiferrago It looks like you brought in some other branches relating to HSTS parsing, and now have a merge conflict. Are you working to fix this? If not, or if you're stuck, don't worry -- I'm happy to resolve the conflict myself and merge in the commits we both :+1:'d in a new PR.

ultiferrago:
@konklone  Yeah haha. I noticed that I jacked it up as I was leaving work. I will be fixing the issue.

ultiferrago:
Fixed the mess I caused with a bad pull. Everything should be good to go now if you approve.

pshtt/utils.py
```diff
@@ -81,7 +81,7 @@ def format_domains(domains):
     formatted_domains = []
 
     for domain in domains:
-        # Replace a single instance of http://, and www. if present.
-        formatted_domains.append(domain.replace("http://", "", 1).replace("www.", "", 1))
+        # Replace a single instance of http://, https://, and www. if present.
+        formatted_domains.append(domain.replace("https://", "").replace("http://", "", 1).replace("www.", "", 1))
```
ultiferrago: If a domain is https://http://www.something.com this would result in something.com. I am not sure if this is okay or not, if not would need to add some logic in order to check for the removal. I could also change the line to use re.sub and a regular expression for replacing one instance of http?://. I thought a regex here might be a tad overkill though.

ultiferrago: re.sub("https?://", "", domain, count=1).replace("www.", "") 

would be the regex method for desired behavior.

michaelblyons: How about combining them and pinning to the string start, so that you don't mangle a (hypothetical) domain like `browncowww.com`?

```python
re.sub("^(https?://)?(www\.)?", "", domain)
```

ultiferrago: Looks good to me. Thoughts @konklone?

konklone: Yes, the regex above perfectly captures the intent of the feature.

konklone:
Tested this locally, and the code works great. Thank you!

PR #77, opened 2017-06-29
Implemented public suffix list.

### LINKED ISSUE #8: Import public suffix list

ultiferrago:


ultiferrago:
public suffix list can be cached but after talking with @h-m-f-t it seems small enough to not need caching (197K) and is only downloaded once per run.

konklone:
> public suffix list can be cached but after talking with @h-m-f-t it seems small enough to not need caching (197K) and is only downloaded once per run.

I think this should still be cacheable, very similar to the preload list JSON we also can download once per run. We do that for the preload list integration so that tools which don't rely on pshtt's built-in batching, but rather perform their own batching around pshtt (such as `domain-scan`), aren't re-downloading the file over and over again.

This does require tools like `domain-scan` to appropriately manage the cached file (and I recently fixed a bug around that in https://github.com/18F/domain-scan/pull/131), but it's pretty necessary to avoid a large number of network requests.

197KiB isn't a lot on its own, but running pshtt 100,000 times to scan 100,000 domains (which is an ordinary volume for GSA's use of domain-scan) would result in 18.7GiB of data transfer, which is a non-trivial cost of time and money.

Could you model the integration of the PSL after the preload-list based approach? 

The [documentation for the `publicsuffixlist` module](https://pypi.python.org/pypi/publicsuffix/) describes how to cache downloaded data, under the `Module content` header.

In a lower section on backwards compatibility, the module encourages users to do this, saying:

> This module ships with an out-dated copy of the Public Suffix List for compatiblity with older versions. Use of this list is deprecated since it encouraged applications that never updated the Public Suffix List data. The built-in list will be removed in a future version.
> 
> If you are using the PublicSuffixList constructor without any arguments, please use the fetch() function and implement an appropriate cache, as shown above.

So that is another reason to download and cache the file locally, to avoid relying on an outdated copy.

h-m-f-t:
@konklone, that only makes sense! Thanks for your help working through this.

konklone:
@ultiferrago 2f7e6a5 looks good! Is this ready for re-review and merge?

ultiferrago:
@konklone Yes I believe it is ready.

konklone:
:+1: Thank you!

PR #79, opened 2017-07-15
Fix crash upon redirects when there's no Location header

### LINKED ISSUE #78: Exception if redirect with no 'Location' header
Working on compliance for lc-idm.llnl.gov I ran into an issue where if the server returns a 3xx (redirect) status, but NOT a 'Location' header, it causes an exception. Adding this check works around that.

The exception is thrown by: https://github.com/dhs-ncats/pshtt/blob/master/pshtt/pshtt.py#L206

A possible fix I have locally is to update the `endpoint.redirect` determination here: https://github.com/dhs-ncats/pshtt/blob/master/pshtt/pshtt.py#L199 to make sure that `req.headers.get('Location') != None`

konklone:
We weren't handling the edge case where a server returns a 3XX status code, but no `Location` header. This was resulting in a crash.

Thanks to @IanLee1521 for reporting this and suggesting the fix. Fixes #78.

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -203,7 +203,8 @@ def basic_check(endpoint):
     endpoint.headers = req.headers
 
     endpoint.status = req.status_code
-    if str(endpoint.status).startswith('3'):
+
+    if (req.headers.get('Location') != None) and str(endpoint.status).startswith('3'):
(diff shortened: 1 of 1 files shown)


h-m-f-t:
👍  thanks @IanLee1521, @konklone

PR #81, opened 2017-07-20
Adding self signed certificate parsing

amirian28:
This PR does three main things:

1) removes "Mozilla NSS CA Store" and replaces it with "Mozilla" 
2) adds parsing for a self-signed certificate
3) adds self-signed-cert key in the https endpoints and the top-level json

### DIFF
--- a/pshtt/models.py
+++ b/pshtt/models.py
@@ -57,6 +57,7 @@ class Endpoint:
         self.https_bad_chain = None
         self.https_bad_hostname = None
         self.https_expired_cert = None
+        self.https_self_signed_cert = None
         self.hsts = None
         self.hsts_header = None
         self.hsts_max_age = None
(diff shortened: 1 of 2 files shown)


amirian28:
Yep! Should be good now. I'll make sure to run that before I submit future PRs. 

So this fixes a few things. 

Re #1 in my original comment: I couldn't get pshtt to verify any certificate previously because it looks like sslyze changed their output, which broke the string parsing. I've changed it to just look for "Mozilla" when appropriate.

Re #2,3: you can try this on self-signed.badssl.com and see the results now in the https endpoints and the top level domain. 

Happy to answer any other questions!

@eric-zeng and @estark37 can you also take a look?

eric-zeng:
Looks good to me!

konklone:
Confirmed that `pki.treas.gov` (which uses an FPKI cert) was coming up with a `HTTPS Bad Chain` value of `False` before applying this PR, and has a value of `True` after applying it.

Also confirmed that iad.gov (which uses a DoD cert from a CA not trusted by any root store and which has a self-signed cert in the chain) now comes up as `True` for having a self-signed cert.

Merging in, thank you!

PR #83, opened 2017-07-21
Use your own CA bundle from the command line

amirian28:
Right now, you can only specify which CA bundle you want to use for requests, and via an environment variable. 

This PR allows you to specific a CA bundle using the command line, and will use the (same) CA bundle for both requests and sslyze. 

This PR also includes modified sslyze output parsing to take this decision into account --- if you use a custom CA bundle, it will only check the output from sslyze for that custom bundle. If you do not, it will continue to look for "Mozilla" as the default.

### DIFF
--- a/pshtt/cli.py
+++ b/pshtt/cli.py
@@ -3,7 +3,7 @@
 """pshtt ("pushed") is a tool to test domains for HTTPS best practices.
 
 Usage:
-  pshtt (INPUT ...) [--output OUTFILE] [--sorted] [--json] [--markdown] [--debug] [--timeout TIMEOUT] [--user-agent AGENT] [--preload-cache PRELOAD] [--cache] [--suffix-cache SUFFIX]
+  pshtt (INPUT ...) [--output OUTFILE] [--sorted] [--json] [--markdown] [--debug] [--timeout TIMEOUT] [--user-agent AGENT] [--preload-cache PRELOAD] [--cache] [--suffix-cache SUFFIX] [--ca-file PATH]
   pshtt (-h | --help)
 
 Options:
@@ -18,6 +18,7 @@ Options:
   -p --preload-cache=PRELOAD  Cache preload list, and where to cache it.
   -c --cache                  Cache network requests to a directory.
   -l --suffix-cache=SUFFIX    Cache suffix list, and where to cache it.
+  -f --ca-file=PATH           Specify custom CA bundle (PEM format)
 
 Notes:
   If the first INPUT ends with .csv, domains will be read from CSV.
@@ -56,7 +57,8 @@ def main():
         'timeout': args['--timeout'],
         'preload_cache': args['--preload-cache'],
         'suffix_cache': args['--suffix-cache'],
-        'cache': args['--cache']
+        'cache': args['--cache'],
+        'ca_file': args['--ca-file']
     }
     results = pshtt.inspect_domains(domains, options)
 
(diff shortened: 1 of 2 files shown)


amirian28:
Forgot to add @eric-zeng and @estark37.  Please take a look!

pshtt/pshtt.py
```diff
@@ -127,6 +131,12 @@ def result_for(domain):
 
 
 def ping(url, allow_redirects=False, verify=True):
+    # If there is a custom CA file and we want to verify
+    # use that instead when pinging with requests
+
+    if CA_FILE and verify:
+        verify = CA_FILE
```
konklone: Is this to trigger the right behavior in `requests`, if you set `verify` to be a path to a `.pem` file rather than a boolean? If so, a comment to that effect would be helpful, as it took me a bit to realize the reason for this assignment.

konklone: I added a comment to that effect in a subsequent commit.

konklone:
Giving @eric-zeng and @estark37 a day or so to ring in if they have any review comments, but this is now a :+1: from me and good to merge.

eric-zeng:
Looks good!

egyptiankarim:
Super great!

konklone:
Resolved the merge conflict, and looks good to go! Thanks @amirian28!

pshtt/pshtt.py
```diff
@@ -127,6 +133,14 @@ def result_for(domain):
 
 
 def ping(url, allow_redirects=False, verify=True):
+    # If there is a custom CA file and we want to verify
+    # use that instead when pinging with requests
```
IanLee1521: Should there be a comma at "... we want to verify, is that ..." ?

pshtt/cli.py
```diff
@@ -56,7 +57,8 @@ def main():
         'timeout': args['--timeout'],
         'preload_cache': args['--preload-cache'],
         'suffix_cache': args['--suffix-cache'],
-        'cache': args['--cache']
+        'cache': args['--cache'],
+        'ca_file': args['--ca-file']
```
IanLee1521: As a minor not, if you add the comma at the end here you won't have to modify this line when you add the next line.

pshtt/pshtt.py
```diff
@@ -127,6 +133,14 @@ def result_for(domain):
...
+
+    # By changing the verify param from a boolean to a .pem file, the
+    # requests module will use the .pem to validate HTTPS connections.
```
IanLee1521: Is `.pem` the only extension allowed? If so, should we check for it?

pshtt/pshtt.py
```diff
@@ -803,9 +867,24 @@ def is_domain_enforces_https(domain):
 
 
 def is_domain_strong_hsts(domain):
+    if is_hsts(domain) and hsts_max_age(domain):
+        return (
+            is_hsts(domain) and
+            hsts_max_age(domain) >= 31536000
+        )
+    else:
+        return None
```
IanLee1521: Should this return `False` instead of `None` ?

requirements.txt
```diff
@@ -5,3 +5,4 @@ docopt
 requests_cache
 pytablewriter
 publicsuffix
+pyopenssl=17.2.0
```
IanLee1521: Is there a reason that pyopenssl is oinned but others are not? If so, should we document that here?

pshtt/pshtt.py
konklone: That's my bad, I added that comment, it should have a comma after `verify`. The `use` word is correct though. I'll fix it in a later commit.

requirements.txt
konklone: This was my addition - you noted a bug with this in #87, which I'll fix (and add a check for to our CI build process).

PR #86, opened 2017-07-23
Explicitly defaults HTTPS validation fields to false instead of null

konklone:
This chips away at a larger issue in `pshtt`, and sets a few HTTPS validation fields explicitly to `False` before checking whether they are `True`.

Builds on #81.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ todo.txt
 *.sqlite
 
 .cache/
+cache/
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -4,6 +4,7 @@ todo.txt
 *.sqlite
 
 .cache/
+cache/
```
konklone: I also added `cache/` as an ignore'd directory as a convenience. (IIRC, `.cache` was also used specifically for an in-library HTTP cache, not user-directed caching.)

konklone:
@h-m-f-t Ready for your review whenever you're back.

h-m-f-t:
💯  looks good!

PR #90, opened 2017-07-27
Syntax fix

egyptiankarim:
Correcting the syntax for version-specific dependency references.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,4 @@ docopt
 requests_cache
 pytablewriter
 publicsuffix
-pyopenssl=17.2.0
+pyopenssl==17.2.0


IanLee1521:
Should it just remove the version number entirely? Is there really a hard requirement on that particular version?

egyptiankarim:
@IanLee1521 my personal preference is usually to unchain dependencies in development branches and lock them down in "published" ones. That said, I didn't go back and add versions to the currently unchained dependencies; I just wanted to fix this because it was breaking my builds.

IanLee1521:
Yeah, I do the same. Just the mix that was getting to me. :)

PR #91, opened 2017-07-27
Redirect error handling

amirian28:
nomadtravel.co.uk causes PSL to error out. This is because their Location in the req headers is incorrect (http:///), which causes the subdomain_immediate to be incorrect, which causes PSL to error out and crash the program. I've added an unknown_error here to handle this and any other websites that might have invalid location headers.

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -241,8 +241,6 @@ def basic_check(endpoint):
     if endpoint.redirect:
         try:
             location_header = req.headers.get('Location')
-            print(req.headers)
-            print(location_header)
             # Absolute redirects (e.g. "https://example.com/Index.aspx")
             if location_header.startswith("http:") or location_header.startswith("https:"):
                 immediate = location_header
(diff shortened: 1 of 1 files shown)


amirian28:
adding @ericzeng and @adrifelt to take a look.

pshtt/pshtt.py
```diff
@@ -239,19 +239,25 @@ def basic_check(endpoint):
...
+            # Relative redirects (e.g. "Location: /Index.aspx").
+            # Construct absolute URI, relative to original request.
+            else:
+                immediate = urlparse.urljoin(endpoint.url, location_header)
+
+            # Chase down the ultimate destination, ignoring any certificate warnings.
+            ultimate_req = None
+        except Exception as err:
+            endpoint.unknown_error = True
+            logging.warn("Unexpected other unknown exception when handling Requests Header.")
```
eric-zeng: nit: "Unhandled exception when...." might be less verbose for these messages

konklone:
The code changes look good, and I confirmed the fix. Thank you, @amirian28!

Issue #88, opened 2017-07-25
Add Python 3.6 classifier


IanLee1521:
I've been using / testing pshtt with Python 3.6 and haven't had any issues. We should update the setup.py classifier list at:

https://github.com/dhs-ncats/pshtt/blob/9ec891f0cbd4865d668309900d782b4030c4572c/setup.py#L50

konklone:
Agreed, and confirmed that Python 3.6.1 works fine.

@h-m-f-t has been handling the PyPi-related work and changes, I'll defer to him for :+1: and implementation.

Issue #87, opened 2017-07-25
requirements.txt error


IanLee1521:
Relates to a comment I left on one of the recent pull requests, but the `requirements.txt` file actually has an error if you try to install it:

```
$ pip install -r requirements.txt
Invalid requirement: 'pyopenssl=17.2.0'
= is not a valid operator. Did you mean == ?
```

I would recommend changing it to simply ``pyopenssl`` without the pinned version.

/cc @konklone

konklone:
Fixed by #90.

PR #92, opened 2017-07-30
Test Python 3.6, and test requirements.txt

konklone:
This updates Travis to run our tests in Python 3.6 and to remove Python 3.4. So this tests on Python 2.7, 3.5, and 3.6. 

It also adds `pip install -r requirements.txt` to the testing commands, to avoid a repeat of #87. 

Also relevant to #88, though #88 is requesting a change in PyPi, not Travis CI.

cc @IanLee1521


konklone:
I'm going to add a commit that intentionally breaks the `requirements.txt`, then undoes the commit, to verify that the Travis build breaks and then fix itself.

konklone:
Er, apologies @h-m-f-t, I meant to hold this open for your review/merge.

.travis.yml
```diff
@@ -4,11 +4,12 @@ sudo: false
 
 python:
   - '2.7'
-  - '3.4'
   - '3.5'
+  - '3.6'
 
 install:
   - pip install flake8
+  - pip install -r requirements.txt
```
IanLee1521: You might also want to test:  `pip install .` before the requirements.txt file to make sure that the setup.py file is up to date. Specifically that it has the install requirements it needs in it.

PR #93, opened 2017-07-30
Added Missing SAN check for cert parsing

amirian28:
Checks and flags a website if it has a missing SAN according to sslyze. You can verify this on arunachaltp.nic.in.

### DIFF
--- a/pshtt/models.py
+++ b/pshtt/models.py
@@ -60,6 +60,7 @@ class Endpoint:
         self.https_bad_hostname = None
         self.https_expired_cert = None
         self.https_self_signed_cert = None
+        self.https_missing_san = None
         self.hsts = None
         self.hsts_header = None
         self.hsts_max_age = None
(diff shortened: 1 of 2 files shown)


amirian28:
@eric-zeng, @estark37 could you LGTM this? Thanks!

konklone:
To me, this feels more like part of the category of "bad hostname", which we already have. That category is meant to be `True` when the part of the validation that fails is hostname validation, which this is a case of.

Detecting it is definitely valuable, and what's in `master` currently doesn't mark Bad Hostname as True. Could you update this PR to set the `https_bad_hostname` flag as False instead of creating a new flag?

amirian28:
Ok, I see where you're coming from --- I had a similar thought to myself earlier today but figured I'd start by separating them out, even though one is kinda a subset of the other. 

So, to clarify, in the sslyze checking in https_check, you'd prefer to see https_bad_hostname set to True  instead of https_missing_san? I'm just a little confused because in your previous comment, you ask to set it to False, hence the clarification.

If that's the case, I can go ahead and make that change and remove the https_missing_san!

konklone:
Sorry, yes, I meant True. :)

amirian28:
Ok, should be fixed now.

estark37:
LGTM!

estark37:
LGTM!

konklone:
Thank you, @amirian28!

PR #96, opened 2017-08-09
Fixed regex

amirian28:
My understanding is that when you have a match with a python regex, it returns an object; if you don't, it returns None. I found this when debugging definitivetechnology.com --- some of the endpoints redirected immediately to www, but the json contradicted itself. 

This PR fixes this problem and correctly sets redirect_immediately_to_www to True when applicable. You can observe this with the httpwww endpoint in definitivetechnology.com

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -286,9 +286,7 @@ def basic_check(endpoint):
             base_immediate = parent_domain_for(subdomain_immediate)
 
             endpoint.redirect_immediately_to = immediate
-            endpoint.redirect_immediately_to_www = re.match(r'^https?://www\.', immediate)
-            if endpoint.redirect_immediately_to_www is not None:
-                endpoint.redirect_immediately_to_www = True
+            endpoint.redirect_immediately_to_www = (re.match(r'^https?://www\.', immediate) is not None)
             endpoint.redirect_immediately_to_https = immediate.startswith("https://")
             endpoint.redirect_immediately_to_http = immediate.startswith("http://")
             endpoint.redirect_immediately_to_external = (base_original != base_immediate)
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -287,6 +287,8 @@ def basic_check(endpoint):
 
             endpoint.redirect_immediately_to = immediate
             endpoint.redirect_immediately_to_www = re.match(r'^https?://www\.', immediate)
+            if endpoint.redirect_immediately_to_www is not None:
+                endpoint.redirect_immediately_to_www = True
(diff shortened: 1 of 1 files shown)


amirian28:
@estark37, @eric-zeng can you LGTM this? TIA.

pshtt/pshtt.py
```diff
@@ -287,6 +287,8 @@ def basic_check(endpoint):
 
             endpoint.redirect_immediately_to = immediate
             endpoint.redirect_immediately_to_www = re.match(r'^https?://www\.', immediate)
+            if endpoint.redirect_immediately_to_www is not None:
+                endpoint.redirect_immediately_to_www = True
```
konklone: what about changing 

```python
endpoint.redirect_immediately_to_www = re.match(r'^https?://www\.', immediate)
```

to:

```
endpoint.redirect_immediately_to_www = (re.match(r'^https?://www\.', immediate) is not None)
```

That seems slightly cleaner to me and more expressive, but I think it works fine either way.

konklone:
Confirmed the fix in behavior, and it corrects the canonical URL to use `www` as intended. Good catch of a bug in our logic! I left a comment with a suggested tweak, but I'm good either way.

Also, curious - @amirian28 it seems unrelated to this PR, but I notice that `pshtt` can't connect to `https://www` (and so the canonical URL is `http://www` instead of `https://www`), but I can get to the HTTPS endpoint in cURL and in my browser. Happy to dig into it if you didn't already spot the reason.

pshtt/pshtt.py
eric-zeng: I agree, I think it would be better not to first store a `MatchObject`  in `redirect_immediately_to_www` and then change the value to a boolean. A slightly less terse alternative could be: 
```
www_match = re.match(r'^https?://www\.', immediate)
endpoint.redirect_immediately_to_www = www_match is not None
```

amirian28:
Alright, that should be fixed! I agree, looks cleaner. 

Re: your other comment, are you talking about definitivetechnology.com? Which domain are you referring to? It seems to be fine on my machine, unless you're talking about a different domain. Happy to help debug on a separate issue/PR.

konklone:
Thank you! :+1:

> Re: your other comment, are you talking about definitivetechnology.com? Which domain are you referring to? It seems to be fine on my machine, unless you're talking about a different domain. Happy to help debug on a separate issue/PR.

Yeah, from both my local laptop and an EC2 server, I get a `live` value of `false` for the `httpswww` endpoint, with no headers recorded. I get that consistently, a few hours ago when I left my comment, and also now. It's interesting if you're seeing something different, but yeah it is definitely a different issue than this one.

PR #97, opened 2017-08-10
Https redirects

amirian28:
This fixes a bug in the https validity checker.

Previously, if a site was live at an https endpoint, it would be flagged as valid https. However, it is possible that site it live but immediately redirect to an http site. If this does not happen, THEN we go ahead and say that the site has valid https. 

This behaviour can be seen on a variety of sites: http://metro.co.uk/, debenhams.com, espn.com.


amirian28:
@eric-zeng and @estark37, can you LGTM this? Thanks!

PR #100, opened 2017-08-24
Adding some whitespace to the debugging output.

egyptiankarim:
The bunched up output when using --debug was difficult to read. This is a simple pull to space things out a bit. I'm going to work a separate pull (eventually) to *really* clean up output, but this should be a serviceable middle step.

### DIFF
--- a/pshtt/cli.py
+++ b/pshtt/cli.py
@@ -69,6 +69,7 @@ def main():
             print(output)
         else:
             utils.write(output, out_file)
+            logging.debug("\n-------------------------\n")
             logging.warn("Wrote results to %s." % out_file)
     # Markdwon can go to STDOUT, or to a file
     elif args['--markdown']:
@@ -85,6 +86,7 @@ def main():
         if args['--output'] is None:
             out_file = 'results.csv'
         pshtt.csv_for(results, out_file)
+        logging.debug("\n-------------------------\n")
         logging.warn("Wrote results to %s." % out_file)
 
 
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -158,7 +158,8 @@ def ping(url, allow_redirects=False, verify=True):
 
 
 def basic_check(endpoint):
-    logging.debug("Pinging %s..." % endpoint.url)
+    logging.debug("\n-------------------------\n")
+    logging.debug("Pinging %s...\n" % endpoint.url)
 
     # Test the endpoint. At first:
     #
@@ -177,7 +178,7 @@ def basic_check(endpoint):
 
     except requests.exceptions.SSLError as err:
         logging.warn("Error validating certificate.")
-        logging.debug("{0}".format(err))
+        logging.debug("{0}\n".format(err))
 
         # Retry with certificate validation disabled.
         try:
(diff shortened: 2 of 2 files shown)


egyptiankarim:
screen shot 2017-08-24 at 09 29 47
Sneak preview!

IanLee1521:
Clarification question, does this still look ok if you add the `--json` flag? I find myself using both a lot.

egyptiankarim:
@IanLee1521 good call! It looked a little cramped. I added a bit more whitespace, and it looks decent now. Thanks!

konklone:
@egyptiankarim Hopefully this isn't a huge PITA, but could we move this stuff to a utils function or two? This is a lot of copy/paste, and another thing for people to remember in adding future sections.

So that'd probably be a `utils.debug()` function which takes the same signature as `logging.debug` but which adds a `\n` at the end, and then either a flag for that function or a new function which adds the dividing line.

egyptiankarim:
@konklone not a PITA at all! This should be better.

egyptiankarim:
That last commit was when I finally realized I was missing an extra line break between method definitions in utils.py 😝

pshtt/cli.py
```diff
@@ -66,6 +66,8 @@ def main():
     if args['--json']:
         output = utils.json_for(results)
         if out_file is None:
+
+            utils.debug("Printing JSON...")
```
konklone: We specifically omit debugging lines when outputting JSON or Markdown to STDOUT, so that downstream clients can pipe the output of `pshtt` directly into other processes that expect properly formatted JSON. Mind removing these new debugging output lines from the JSON and Markdown emitters?

pshtt/utils.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
```
konklone: I think this new hashbang line is not necessary, and could be confusing. (I think the file is also not executable, so it won't necessarily be usable without that anyway.)

pshtt/utils.py
```diff
@@ -86,3 +88,11 @@ def format_domains(domains):
...
 
     return formatted_domains
+
+
+def debug(message, divider=True):
+    if divider:
+        logging.debug("\n-------------------------\n")
+
+    if message:
+        logging.debug("%s\n" % message)
```
konklone: Sorry to be a nit, but looking at the usage in `pshtt.py`, it looks like `divider=False` would be a better default and less surprising. That way `utils.debug("Message")` just outputs a message and doesn't cause an additional line to appear.

Also, when referencing this in `pshtt.py`, it would be more explicit and readable to reference the named argument directly. So for example (assuming a reversed default), a call might look like this:

```python
utils.debug("My message!", divider=True)
```

And to do it with no divider:

```python
utils.debug("My message!")
```

egyptiankarim:
@konklone that makes sense to me! Latest commit changes the defaults and modifies messages throughout pshtt.py and cli.py to match convention.

pshtt/pshtt.py
```diff
@@ -177,31 +177,31 @@ def basic_check(endpoint):
 
     except requests.exceptions.SSLError as err:
         logging.warn("Error validating certificate.")
-        utils.debug("{0}".format(err), False)
+        utils.debug("{0}".format(err))
 
         # Retry with certificate validation disabled.
         try:
-            req = ping(endpoint.url, verify=False)
+            req = ping(endpoint.url)
```
konklone: This change looks unintentional and like a bug. Regex gone awry?

konklone:
@egyptiankarim Thanks! I spotted one bug that was just introduced in the last commit. If that's fixed, a :+1: from me for merge.

konklone:
Also, it would be good to get @h-m-f-t's :+1: on the output changing in this way, and the implementation, before merging.

pshtt/pshtt.py
egyptiankarim: Gah! Yes! Excellent catch.

konklone:
Thanks, verified the fix. Will defer to @h-m-f-t for the :+1: and merge.

h-m-f-t:
LGTM @egyptiankarim 👍

PR #101, opened 2017-08-25
add a script for running via Docker

afeld:
This simplifies running with Docker, hiding the complexity of the commands from the user.


egyptiankarim:
Would it make sense to merge my proposed changes from #103 into this PR? I can take care of it if so.

afeld:
I think they can be reviewed/merged independently.

PR #103, opened 2017-08-28
Simplifying Docker usage of custom trust stores

egyptiankarim:
Just making a few changes that allow us to treat the PEM files that contain our custom trust stores like we do other `pshtt` inputs. Specifically, I'm copying those PEM files to the Docker /usr/src/app working directory and telling gitignore not to watch them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 todo.txt
 *.json
 *.sqlite
+*.pem
 
 .cache/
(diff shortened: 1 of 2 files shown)


konklone:
:+1: Seems reasonable and low-risk.

konklone:
:+1: Seems reasonable and low-risk.

PR #104, opened 2017-09-05
Added exception handling for network requests

IanLee1521:
This is necessary if you are running a scan from inside a network where you can't reach the internet resources and / or where there is SSL MITM being performed on the requests, causing SSL errors.


IanLee1521:
Yup, will get that fixed this afternoon.

konklone:
Good with me, deferring to @h-m-f-t for merge.

h-m-f-t:
Thanks for this addition, @IanLee1521.

PR #108, opened 2017-09-15
Added .xls and .xlsx to gitignore

IanLee1521:


konklone:
👍 from me - though it seems like good hygiene to at least have a nominal description of the reason for the change in the issue. 

I assume you're using `pshtt`'s built-in CSV output and then turning them into XLS files within the project directory?

IanLee1521:
That's a good point, yes that's exactly right. I'm turning them into filtered / colored reports in .xlsx

PR #110, opened 2017-09-18
Bumps version number to next dev version

IanLee1521:
Also moves to 0.3.x rather than 0.2.x line, to match Semantic Versioning (semver.org) philosophies.

Adding the `-dev` allows a user working with the git repo directly to know that they are using a different version than the "last official release".


konklone:
Defer to @h-m-f-t for PyPi-related changes, but :+1: from me.

Though FWIW, I think 0.x has very few rules. From semver.org:

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

IanLee1521:
It's true, I just try to reserve bump the minor version on any forward or backward in/compatibility changes. Gets in a good habit for once we get to `1.y.z`

PR #112, opened 2017-09-20
Updated code to move docstrings into functions

IanLee1521:
This allow the docstrings to be accessible to Python directly, and better utilizing the docstring conventions from: https://www.python.org/dev/peps/pep-0257/


konklone:
:+1:, and thank you!

PR #113, opened 2017-09-20
Default to building universal wheel

IanLee1521:
Because the project is pure Python and 2/3 compatible,
we can build a single wheel out of the package using:

`$ python setup.py bdist_wheel`

Ref: https://packaging.python.org/tutorials/distributing-packages/#universal-wheels

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,5 @@
+[bdist_wheel]
+universal = true
+
 [metadata]
 description-file = README.md


konklone:
Are you sure? I think `sslyze` has C extensions.

IanLee1521:
It does, but the "universal-ness" of a package depends only on itself, not on its dependencies.

konklone:
OK, that seems reasonable (and yeah, rereading the docs, I understand better now).

The examples in the docs say `universal = 1`, but you used `universal = true`. Will it work either way?

IanLee1521:
Yes, I've used true in other places (and think it's clearer / better documenting). 

I also tested building the wheel locally.

konklone:
Cool - looks good to me, but deferring to @h-m-f-t for merges of PyPi-related stuff, since I don't have any role in the packaging process (and haven't ever managed a PyPi package).

PR #114, opened 2017-09-22
Scripts and Documentation on how to run Pshtt on multiple GCE Instances

amirian28:
It is possible that an individual would like to run Pshtt at a large scale (1 million or more) and doesn't have internal infrastructure to do this. For about $100-200, a user can use these scripts to run Pshtt on Google Cloud Engine on 1 Million sites.

The following PR has an array of scripts that will help a user run pshtt at a large scale using GCE.


gce-scripts/README.md
```diff
@@ -0,0 +1,205 @@
...
+    *   `sudo ./packages_to_install.sh`
+2.  Clone pshtt.
+    *   `git clone https://github.com/dhs-ncats/pshtt.git`
+3.  Put roots.pem, running_script.sh, and your input file in the same dir as
+    pshtt.
+    *   Follow directions under Updating data files above on how to get a
+        roots.pem.
+    *   domains must have the schema stripped of them and no trailing '/', such
+        as:
+        *   domain.tld/path/to/page
```
konklone: `domain.tld/path/to/page` isn't a supported thing to do with `pshtt`, and unfortunately right now will even screw up the base domain calculator.

gce-scripts/packages_to_install.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+# Installs all the necessary packages for pshtt to run.
+# Logs which package it is installing as well as it's success (0) or failure
+# (1).
+echo 'UPDATE'
+apt-get -y update -qq
+echo $? ' ERROR CODE'
+echo 'PYTHON PIP'
+apt-get -y install python-pip -qq
```
konklone: Is `python-pip` required in addition to `python3-pip` below? I don't think we have any Python 2 dependencies anymore.

gce-scripts/packages_to_install.sh
```diff
@@ -0,0 +1,51 @@
...
+apt-get -y install git -qq
+echo $? ' ERROR CODE'
+echo 'PYTHON3-PIP'
+apt-get -y install python3-pip -qq
+echo $? ' ERROR CODE'
+echo 'LIBFFI6'
+apt-get -y install libffi6 libffi-dev -qq
+echo $? ' ERROR CODE'
+echo 'LIBSSL'
+apt-get -y install build-essential libssl-dev libffi-dev python-dev python3-dev -qq
```
konklone: Same question here with `python-dev` and `python3-dev`.

gce-scripts/running_script.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Runs pshtt with a 10 second time, "Google-Transparency-Report" as the user
```
konklone: This cites `"Google-Transparency-Report"` as the user agent, but this may not be intended, and I don't see this string mentioned anywhere else in the PR.

konklone: Also I think this is meant to be `10 second timeout` rather than `10 second time`?

konklone:
LGTM - @h-m-f-t, you a :+1: to merge?

gce-scripts/combine_shards.py
```diff
@@ -0,0 +1,29 @@
...
+import sys
+
+
+def main():
+    if (len(sys.argv)) < 2:
+        print('you need a filename!')
+        exit(1)
+    # Master file is the file with the list of filenames to intake.
+    # Fileception.
+    master_file = sys.argv[1]
```
IanLee1521: May not be, but is it necessary to do any error handling here? Either check the file exists or catch the exception?

gce-scripts/packages_to_install.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+# Installs all the necessary packages for pshtt to run.
+# Logs which package it is installing as well as it's success (0) or failure
+# (1).
+echo 'UPDATE'
```
IanLee1521: Rather than all the `echo ...` 's, would it be sufficient to make use of `set -x` bash command?

gce-scripts/packages_to_install.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+# Installs all the necessary packages for pshtt to run.
+# Logs which package it is installing as well as it's success (0) or failure
```
IanLee1521: Is failure OK in this case or should the script error out if a package install fails? If so, perhaps it would be simpler to use `set -e` at the top of the file?

gce-scripts/running_script.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Runs pshtt with a 10 second timeout, with roots.pem as the CA file,
+# and debug on. Logging goes to time_<input_file_name>.txt
+
+# ./running_script.sh test_file.csv
+# output files: test_file.csv.json, time_test_file.csv.txt
+
+input_file=$1
```
IanLee1521: Might be worth doing a check to make sure there was an argument provided?

gce-scripts/scp_and_setup.sh
```diff
@@ -0,0 +1,139 @@
...
+    # uploading.
+    # If we are uploading file #3 in the list, go to line 3 in the hosts file
+    # and upload to that ip.
+
+    machine=$(sed "${i}q;d" $hosts_file)
+    echo 'Now on '"${machine}"' number '$i
+    # Do not do strict host key checking so that you dont have to type "yes" for
+    # each machine.
+    echo 'Uploading packages_to_install.sh'
+    scp -i ~/.ssh/gce_pshtt_key -o "StrictHostKeyChecking no" packages_to_install.sh ubuntu@"${machine}":~/
```
IanLee1521: Is there a reason to do `-o "StrictHostKeyChecking no"` here, but not in the other commands? I see there is the comment above, but will it re-prompt at the next call?

gce-scripts/scp_and_setup.sh
```diff
@@ -0,0 +1,139 @@
...
+    echo $?
+    # We echo after each command to ensure that it worked. 0 means success.
+    # The Log file is how we can tell if the packages have all been uploaded.
+    echo 'Creating packages log file'
+    ssh -i ~/.ssh/gce_pshtt_key ubuntu@"${machine}" touch package_log_file.txt
```
IanLee1521: Would it be preferable to make a file level alias up above to factor out the ssh requirements? Something like `alias _ssh='ssh -i ~/.ssh/gce_pshtt_key -o "StrictHostKeyChecking no" ubuntu@"${machine}"'` which could then simplify this and other lines to: `_ssh touch package_log_file.txt`

gce-scripts/scp_and_setup.sh
```diff
@@ -0,0 +1,139 @@
...
+i=1
+for y in $list_of_files;
+do
+    machine=$(sed "${i}q;d" $hosts_file)
+    echo 'Now on '"${machine}"' number '$i
+    echo 'Cloning github repo file'
+    ssh -i ~/.ssh/gce_pshtt_key -t ubuntu@"${machine}" git clone https://github.com/dhs-ncats/pshtt.git
+    echo $?
+    echo 'copying data file to pshtt directory'
+    scp -i ~/.ssh/gce_pshtt_key input_files/"${y}" ubuntu@"${machine}":~/pshtt/
```
IanLee1521: Similar to the comment above about an `_ssh` alias, maybe there could be a similar command for `alias _scp='scp -i ~/.ssh/gce_pshtt_key -o "StrictHostKeyChecking no" '`

gce-scripts/run_all_scripts.sh
```diff
@@ -0,0 +1,49 @@
...
+# alexa_. So the shard files will be alexa000.csv, alexa001.csv
+# etc.
+
+# If any of the scripts fails, this hard fails and tells the user what script
+# went wrong.
+
+
+input_file=$1
+number_of_shards=${2-10}
+output_file_name=${3-shard_}
```
IanLee1521: Maybe I'm missing it, but what is `shard_` ? I don't see it set anyplace else...

h-m-f-t:
Thanks @amirian28! @dhs-ncats won't give me a few extra Benjamins to actually try this out on GCE, but I'm happy to add this documentation into the repo and I'm glad that you would share it upstream. 🎉

amirian28:
Thanks! And sorry @IanLee1521 for not addressing your comments --- work got unexpectedly hectic and this PR slipped my mind. Those are great comments/suggestions, so I'm hoping to go back and fix them up in the near future. I'm no bash expert, so the comments are appreciated.

IanLee1521:
No worries, happens to us all. :)

As for the comments, I'm no bash expert either, so happy to help out and continue the conversation on the next pull request. :)

PR #115, opened 2017-09-26
Add self-signed errors to bad chain calculation

h-m-f-t:
This is a small change with cascading effects: treating self-signed certificates in the certificate chain as "bad chain" errors. It is proposed to more accurately represent what constitutes bad chain errors, and also to more correctly score domains that "support HTTPS".

"Domain Supports HTTPS" is an adjudication of several factors. Even if a domain is "true" for "HTTPS Bad Chain", pshtt allows "Domain Supports HTTPS" to be true in recognition of the fact that a domain with certificate validation challenges does still _support_ the use of HTTPS. This change scores "HTTPS Bad Chain" as true when there is a self-signed certificate in the chain, which allows these domains to be scored as "supporting HTTPS".

This change does mean that any time a domain is "HTTPS Self Signed Cert"=="True" that "HTTPS Bad Chain" is true also. We may consider collapsing these fields into one, though I believe having a separate output for self-signed certs is helpful.

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -494,8 +494,10 @@ def https_check(endpoint):
         if (
             (STORE in msg) and
             (("FAILED") in msg) and
-            ((("unable to get local issuer certificate") in msg) or
-            (("self signed certificate") in msg))
+            (
+                (("unable to get local issuer certificate") in msg) or
+                (("self signed certificate") in msg)
+            )
         ):
             endpoint.https_bad_chain = True
 
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -494,7 +494,8 @@ def https_check(endpoint):
         if (
             (STORE in msg) and
             (("FAILED") in msg) and
-            (("unable to get local issuer certificate") in msg)
+            ((("unable to get local issuer certificate") in msg) or
+            (("self signed certificate") in msg))
         ):
             endpoint.https_bad_chain = True
 


konklone:
This is an elegant way of handling this (given the baseline that we're using inelegant text parsing from sslyze output to begin with!), and I know this issue needs to be resolved soon. I'm a :+1: to merge as an interim, but would advocate for a broader solution (and am happy to work on one).

> ...that a domain with certificate validation challenges does still support the use of HTTPS.

What I've been trying to aim for in some of these past decisions is to say that using a non-publicly-trusted root CA is still potentially supporting the use of HTTPS. Using private CAs or non-Mozilla-trusted CAs can be done securely for the right set of clients.

Using a self-signed certificate is very rarely able to be done securely -- only for very specialized situations with novel clients that perform validation in some other off-band way. And even then that's theoretical for me, I've never yet seen this done in practice.

I know we're experiencing large false positives with some agency services - is that because sslyze is calling something a self-signed cert when it's actually an FPKI cert it can't validate?

If we can't reasonably distinguish actually self-signed certs from bad chains, then merging their detection is the only reasonable call, though it's too bad if that's the case.

> We may consider collapsing these fields into one, though I believe having a separate output for self-signed certs is helpful.

Yes, it's definitely helpful to keep those as separate -- though, only if it's a reasonably accurate distinction. If a bunch of services are causing Self-Signed to register as True, but they're actually just using a non-Mozilla-trusted CA (and didn't include an intermediate in their chain), then that extra field won't help them any.

So that's all to say - :+1: and I'll merge, though I'd like to dig in a little further in discussion about how/if we can reliably distinguish self-signed certs at all. If we can't, perhaps we should remove the field altogether, or pursue another method besides sslyze output for detecting those issues.

konklone:
(Pushed a commit to fix the `flake8` warning.)

PR #118, opened 2017-09-30
Converted model classes to new style

IanLee1521:
This is a Python 2 -> 3 difference. Reference: https://www.python.org/doc/newstyle/


konklone:
:+1:

PR #121, opened 2017-09-30
Updated Python versions support

IanLee1521:
Travis was testing Python 3.6, not included in the setup.py
classifiers list. Meanwhile, 3.4 is in the classifiers, but
not being tested. This brings those two in sync.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ sudo: false
 
 python:
   - '2.7'
+  - '3.4'
   - '3.5'
   - '3.6'
 
@@ -13,4 +14,3 @@ install:
 
 script:
   - flake8 .
-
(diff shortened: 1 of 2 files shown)


konklone:
I think we can probably drop 3.4 from both, personally. But :+1: from me either way. @h-m-f-t ?

IanLee1521:
Possibly.

My feeling on this has often been that "as long as I don't have to do anything more / new / extra to "support" an older version, I will. The minute I go to add a feature that wouldn't work in Python3.4 (for instance) is when I would think about dropping it.

Based on the tests though, this works in Python 3.4 with no changes needed, so I vote we keep it. (Also, since I think Python3.4 is/was the default Python3 in a few versions of Linux, e.g. Ubuntu 14.04, etc)

Issue #117, opened 2017-09-29
Add code coverage GitHub integration


IanLee1521:
https://coveralls.io

Related to my work in: https://github.com/ianlee1521/pshtt/tree/test-framework

IanLee1521:
Hmm. Or maybe codecov.io would be better?

https://www.slant.co/improve/versus/7928/7929/~coveralls_vs_codecov

IanLee1521:
Example in that fork where I just got this to track the results: https://coveralls.io/github/IanLee1521/pshtt

IanLee1521:
Defer to @h-m-f-t, as enabling this requires repository admin rights.

h-m-f-t:
Resolved with #119.

PR #122, opened 2017-10-02
Test framework

IanLee1521:
This pull request adds a test harness for pshtt which enables functional testing, and begins work (skipped for now: https://github.com/dhs-ncats/pshtt/compare/master...IanLee1521:test-framework#diff-4b8ef9663ddcb745c9b2af6c8974c69fR32) to allow for integration tests against live badssl.com endpoints. 

Instead of live hits to badssl though, it may be preferable to record the HTTP requests/responses and replay those, e.g. using [`vcrpy`](https://pypi.python.org/pypi/vcrpy), investigation of which is still on going and will be added in future pull requests.


IanLee1521:
The failure here is due to a need for an admin ( @h-m-f-t ) to set up the integration with Coveralls.io (see #117)

PR #125, opened 2017-10-21
Re-architecting of CSV / MD / JSON writing

### LINKED ISSUE #28: Output a row to the CSV after each scan
And catch any crashes with a generic exception handler _just_ to close the CSV file, to ensure that any partial scan results still get written out in the case of a crash.

IanLee1521:
This resolves #28, by re-working the cli main function to write out each domain of data as it is generated. This works by converting the `inspect_domains` function yield the results of each domain inspection as they occur, rather than waiting until the very end of the generation.

This PR also simplifies the logic surrounding writing out the other formats (JSON and MD) by using a unified `smart_open` context manager to manage outputting to a file vs to stdout.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -41,9 +41,9 @@ class TestToCSV(unittest.TestCase):
         with open(self.temp_filename) as fh:
             content = fh.read()
 
-            expected = 'Domain,Base Domain,Canonical URL,Live,Redirect,Redirect To,Valid HTTPS,Defaults to HTTPS,Downgrades HTTPS,Strictly Forces HTTPS,HTTPS Bad Chain,HTTPS Bad Hostname,HTTPS Expired Cert,HTTPS Self Signed Cert,HSTS,HSTS Header,HSTS Max Age,HSTS Entire Domain,HSTS Preload Ready,HSTS Preload Pending,HSTS Preloaded,Base Domain HSTS Preloaded,Domain Supports HTTPS,Domain Enforces HTTPS,Domain Uses Strong HSTS,Unknown Error\n'
+        expected = ','.join(_pshtt.HEADERS) + '\n'
 
-            self.assertEqual(content, expected)
+        self.assertEqual(content, expected)
 
     @unittest.skipIf(sys.version_info[0] < 3, 'Python 3 test only')
     def test_single_result(self):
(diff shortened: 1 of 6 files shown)


IanLee1521:
@h-m-f-t , it looks like you may have left coveralls configured to require only increases in coverage, but since this removes some code, it actually dropped the coverage.

Until we can get the testing framework more intact, maybe we should bump up the threshold for a pull request dropping the coverage to, maybe something like, 2%? 5%?

Should be able to be configured at: https://coveralls.io/github/dhs-ncats/pshtt/settings (though you as admin would need to do this)

IanLee1521:
@h-m-f-t , For this particular case, I ended up just updating the pull request to add more tests along the way. :)

IanLee1521:
Alright, I know I was doing a bunch of work after submitting this, but I'm done for a while. The last thing I would like to do with this is to add the testing of the `to_json` and `to_markdown` functions I factored out, but I can do that in this PR or a separate one if we get this merged sooner.

tests/test_cli.py
```diff
@@ -0,0 +1,59 @@
...
+        cls.temp_filename = os.path.join(tempfile.gettempdir(), 'results.csv')
+
+    @unittest.skipIf(sys.version_info[0] < 3, 'Python 3 test only')
+    def test_no_results(self):
+        to_csv([], self.temp_filename)
+
+        with open(self.temp_filename) as fh:
+            content = fh.read()
+
+            expected = 'Domain,Base Domain,Canonical URL,Live,Redirect,Redirect To,Valid HTTPS,Defaults to HTTPS,Downgrades HTTPS,Strictly Forces HTTPS,HTTPS Bad Chain,HTTPS Bad Hostname,HTTPS Expired Cert,HTTPS Self Signed Cert,HSTS,HSTS Header,HSTS Max Age,HSTS Entire Domain,HSTS Preload Ready,HSTS Preload Pending,HSTS Preloaded,Base Domain HSTS Preloaded,Domain Supports HTTPS,Domain Enforces HTTPS,Domain Uses Strong HSTS,Unknown Error\n'
```
konklone: Could we generate this by pulling in the global var and joining them with `.`? That would prevent us from having to update the test data every time a column header changes.

tests/test_cli.py
```diff
@@ -0,0 +1,59 @@
...
+    @unittest.skipIf(sys.version_info[0] < 3, 'Python 3 test only')
+    def test_single_result(self):
+        to_csv([self.results], self.temp_filename)
+
+        with open(self.temp_filename) as fh:
+            content = fh.read()
+
+            expected = ''
+            expected += 'Domain,Base Domain,Canonical URL,Live,Redirect,Redirect To,Valid HTTPS,Defaults to HTTPS,Downgrades HTTPS,Strictly Forces HTTPS,HTTPS Bad Chain,HTTPS Bad Hostname,HTTPS Expired Cert,HTTPS Self Signed Cert,HSTS,HSTS Header,HSTS Max Age,HSTS Entire Domain,HSTS Preload Ready,HSTS Preload Pending,HSTS Preloaded,Base Domain HSTS Preloaded,Domain Supports HTTPS,Domain Enforces HTTPS,Domain Uses Strong HSTS,Unknown Error\n'
+            expected += 'example.com,example.com,http://example.com,False,False,,False,False,False,False,False,False,False,False,False,,,False,False,False,False,False,False,False,False,False\n'
```
konklone: I think this would be best managed in the test data as an array which is also joined by a comma in code before appending to `expected`. It still might require changing when the headers change, but it would be much more obvious and direct how to make that change.

konklone:
Side question - @h-m-f-t, the Coveralls integration is a bit noisy. Is there a way to tune it down somewhat?

tests/test_cli.py
IanLee1521: (Replying to all three comments) Yeah, I actually started down that route originally. For the headers, as you say, using `','.join(_pshtt.HEADERS)` is pretty simple, the issue for me then became how to convert the data itself, as doing a list of the values will make it tougher to map each back to the appropriate header...

Maybe creating a full domain object and setting all the values there would be best... Let me give that a shot.

IanLee1521:
The coveralls output might be more my fault... It fires on every push, and since I'd kept working on this PR after creating it (which I probably shouldn't have) that's why it got a bit noisy).

IanLee1521:
Huh, I don't see those... this is what I see:

```bash
[... snip ...]
flake8 installed: asn1crypto==0.23.0,beautifulsoup4==4.6.0,certifi==2017.7.27.1,cffi==1.11.2,chardet==3.0.4,cryptography==1.9,DataProperty==0.25.6,docopt==0.6.2,dominate==2.3.1,elasticsearch==5.4.0,flake8==3.4.1,idna==2.6,jsonschema==2.6.0,Logbook==1.1.0,markdown2==2.3.4,mbstrdecoder==0.2.2,mccabe==0.6.1,nassl==0.17.0,path.py==10.4,pathvalidate==0.16.2,pshtt==0.3.0.dev0,publicsuffix==1.1.0,pycodestyle==2.3.1,pycparser==2.18,pyflakes==1.5.0,pyOpenSSL==17.3.0,pyparsing==2.2.0,pytablereader==0.13.4,pytablewriter==0.24.0,python-dateutil==2.6.1,pytz==2017.2,requests==2.18.4,requests-cache==0.4.13,SimpleSQLite==0.16.0,six==1.11.0,SSLyze==1.1.4,tls-parser==1.1.0,toml==0.9.3,typepy==0.0.20,urllib3==1.22,wget==3.2,xlrd==1.1.0,XlsxWriter==1.0.2,xlwt==1.3.0
flake8 runtests: PYTHONHASHSEED='4263820095'
flake8 runtests: commands[0] | flake8
___________________________________________________________________________________ summary ____________________________________________________________________________________
  py27: commands succeeded
SKIPPED:  py34: InterpreterNotFound: python3.4
SKIPPED:  py35: InterpreterNotFound: python3.5
  py36: commands succeeded
  flake8: commands succeeded
  congratulations :)
```

Can you paste what you're seeing?

konklone:
> The coveralls output might be more my fault... It fires on every push, and since I'd kept working on this PR after creating it (which I probably shouldn't have) that's why it got a bit noisy).

Absolutely not -- filing a PR and then continuing to modify it in-place is a design feature of GitHub and a positive part of collaborative development. Please keep doing that! We should just make coveralls less noisy so as not to punish that workflow.

tests/test_cli.py
konklone: >  as doing a list of the values will make it tougher to map each back to the appropriate header...

This is true, but it's probably an acceptable annoyance, IMO. I think it's likely to be less brittle than a concatenated string, and so just turning that into a static array would be enough for me for this PR.

jsf9k:
@IanLee1521, [here](https://github.com/dhs-ncats/pshtt/files/1410794/tox.txt) is the output I get from `tox`.

IanLee1521:
@jsf9k -- It looks like the issue is that you have a virtualenv directory in your top level directory. Therefore flake8 is picking up those files... One way to solve this is #127. Does that work for you?

konklone:
> @konklone does https://github.com/dhs-ncats/pshtt/pull/125/files#diff-09f983a1db5a7aca6102f2e2a260c88cR55 work better for you w.r.t. the static string versus something better?

That works (I also would have been fine with a straight array with header names in the comments, but all good).

PR #129, opened 2017-10-30
Correct load_domains behavior for skipping rows

ericlaw1979:
Ignore empty lines and only ignore a "header" if it's the first entry in the file.


jsf9k:
Looks good to me.  Thanks @ericlaw1979!

PR #131, opened 2017-11-05
Use one flag to cache all third party data in one directory

### LINKED ISSUE #99: Being nicer to hstspreload.org, chromium.googlesource.com, and publicsuffix.org

konklone:
This removes the individual flags for the preload list and PSL, and makes one new flag that covers the preload list, PSL, and the preload pending list.

In addition to making @lgarron and the hstspreload.org servers a lot happier, this will make it easier for people/tools that make use of `pshtt` to rationalize their caching strategy, and easier for the `pshtt` maintainers to add new third party data sources without adding new flags and approaches to caching.

Fixes #99.


konklone:
Also, I should note -- this removes the older flags (`--suffix-cache and `preload-cache`), and so people making use of those two flags will find that caching no longer functions. However, I think I am the only one making use of it! And it seemed worth clearing up flag cruft now, while the tool is still young. So I will plan to make a coordinated update on my end once a new version of `pshtt` is published with this addition, assuming it's merged.

pshtt/cli.py
```diff
@@ -3,22 +3,20 @@
...
+  -h --help                     Show this message.
+  -s --sorted                   Sort output by domain, A-Z.
+  -o --output=OUTFILE           Name output file. (Defaults to "results".)
+  -j --json                     Get results in JSON. (Defaults to CSV.)
+  -m --markdown                 Get results in Markdown. (Defaults to CSV.)
+  -d --debug                    Print debug output.
+  -u --user-agent=AGENT         Override user agent.
+  -t --timeout=TIMEOUT          Override timeout (in seconds).
+  -c --cache-third-parties=DIR  Cache third party data, and what directory to cache it in.
+  -f --ca-file=PATH             Specify custom CA bundle (PEM format)
```
konklone: Some of the changes here are from syncing the options in `cli.py` with the options documented in the README, which required making changes in both directions.

pshtt/cli.py
```diff
@@ -3,22 +3,20 @@
 """pshtt ("pushed") is a tool to test domains for HTTPS best practices.
 
 Usage:
-  pshtt (INPUT ...) [--output OUTFILE] [--sorted] [--json] [--markdown] [--debug] [--timeout TIMEOUT] [--user-agent AGENT] [--preload-cache PRELOAD] [--cache] [--suffix-cache SUFFIX] [--ca-file PATH]
+  pshtt (INPUT ...) [--output OUTFILE] [--sorted] [--json] [--markdown] [--debug] [--timeout TIMEOUT] [--user-agent AGENT] [--cache-third-parties DIR] [--ca-file PATH]
```
konklone: I also synced the options between the `Usage:` line here and the list below, as they were slightly out of sync.

pshtt/cli.py
```diff
@@ -107,9 +105,7 @@ def main():
     options = {
         'user_agent': args['--user-agent'],
         'timeout': args['--timeout'],
-        'preload_cache': args['--preload-cache'],
-        'suffix_cache': args['--suffix-cache'],
-        'cache': args['--cache'],
```
konklone: The `--cache` flag has been unused for a long time -- it's a remnant of a time when this library used `requests-cache`, but that was removed a long time ago.

pshtt/pshtt.py
```diff
@@ -53,14 +50,24 @@
...
+
+# Used for caching the HSTS pending preload list from hstspreload.org.
+PRELOAD_PENDING_CACHE = None
+PRELOAD_PENDING_CACHE_DEFAULT = "preload-pending.json"
 preload_pending = None
 
 # Used for determining base domain via Mozilla's public suffix list.
-SUFFIX_CACHE = None
+PUBLIC_SUFFIX_CACHE = None
+PUBLIC_SUFFIX_CACHE_DEFAULT = "public-suffix-list.txt"
```
konklone: I didn't provide a way to override the individual filenames within the third party cache directory. Though options are always nice, it doesn't seem like it has enough utility to merit the complexity.

pshtt/pshtt.py
```diff
@@ -1015,28 +1022,38 @@ def did_domain_error(domain):
     )
 
 
-def fetch_preload_pending():
+def load_preload_pending():
```
konklone: Standardized on `load_` as the prefix for these 3 loading/caching functions, since they were split among `load_`, `create_`, and `fetch_` before, which seemed slightly confusing.

pshtt/cli.py
konklone: For info on that, see https://github.com/reclosedev/requests-cache/issues/70.

konklone:
Thanks for the review, @jsf9k! 

Next up is a PR letting third party cache data be passed in via variable, to support environments like Lambda where you can reuse in-memory data across containers.

lgarron:
> Next up is a PR letting third party cache data be passed in via variable, to support environments like Lambda where you can reuse in-memory data across containers.

`$XDG_CACHE_HOME`, plz? :-D

PR #136, opened 2017-11-11
Python API for advanced third party caching

konklone:
This adds support for Python API users passing in list values as the cache values for third party downloads. This allows users of `pshtt` in novel environments to have greater control over how and when caching is performed.

This PR also adjusts `initialize_external_data()` so that if the lists have already been downloaded or set once, running it 2 or more times won't change the output (no matter what's passed in after the 1st time). This lets users call it before asking `pshtt` to inspect domains, while still letting `pshtt` call it itself. I could see changing this to have `pshtt` examine whether third parties are already set, but this way also has the property of allowing the third parties to be selectively set in Python (for example, setting 2 out of 3 of them) and then having the call inside `inspect_domains` fill in any remaining gaps.

It also implicitly advertises the three `load_*` third party functions as part of the Python API to the tool, which I am also using in `domain-scan`.

For an example of this in use, see this almost-merged content in `domain-scan`:
https://github.com/18F/domain-scan/blob/94a749f59242bb22aab46d5f2271a36c4fe1e3eb/scanners/pshtt.py

That code downloads two of the third party data sources locally and then chops them down to the smallest possible subset before transmitting them over the network to Lambda, which is where the call to `inspect_domains` is made. The third data source (PSL) is a `.dat` file, which can't be serialized as easily, so it's downloaded inside Lambda explicitly, but is cached in a local variable where Lambda container reuse can make a best-effort approach at retaining that cached data (generally, once per worker thread, so 100 worker threads for 1300 domains would likely download it 100 times total).

Though the use case is a bit specialized, it's added flexibility without much additional code complexity -- and since third party downloads are a sensitive code path (since using it incorrectly could DoS a helpful public service) it seems appropriate to provide as much flexibility as possible.

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -1089,6 +1089,8 @@ def load_preload_list():
     return fully_preloaded
 
 
+# Returns an instantiated PublicSuffixList object, and the
+# list of lines read from the file.
 def load_suffix_list():
     # File does not exist, download current list and cache it at given location.
     utils.debug("Downloading the Public Suffix List...", divider=True)
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -51,18 +51,15 @@ HEADERS = [
 ]
 
 # Used for caching the HSTS preload list from Chromium's source.
-PRELOAD_CACHE = None
-PRELOAD_CACHE_DEFAULT = "preloaded.json"
+cache_preload_list_default = "preloaded.json"
 preload_list = None
 
 # Used for caching the HSTS pending preload list from hstspreload.org.
-PRELOAD_PENDING_CACHE = None
-PRELOAD_PENDING_CACHE_DEFAULT = "preload-pending.json"
+cache_preload_pending_default = "preload-pending.json"
 preload_pending = None
 
 # Used for determining base domain via Mozilla's public suffix list.
-PUBLIC_SUFFIX_CACHE = None
-PUBLIC_SUFFIX_CACHE_DEFAULT = "public-suffix-list.txt"
+cache_suffix_list_default = "public-suffix-list.txt"
 suffix_list = None
 
 # Directory to cache all third party responses, if set by user.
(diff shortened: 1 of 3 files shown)


pshtt/pshtt.py
```diff
@@ -1134,30 +1134,46 @@ def load_suffix_list():
...
+    cached data in a customized way.
+
+    It also potentially allows clients to pass in subsets of these lists,
+    for testing or novel performance reasons.
+
+    Otherwise, if the --cache-third-parties=[DIR] flag specifies a directory,
+    all downloaded third party data will be cached in a directory, and
+    used from cache on the next pshtt run instead of hitting the network.
+
+    If no values are passed in, and no --cache-third-parties flag is used,
```
IanLee1521: @konklone -- Should the caching / interaction with this flag take place in this function? Just to keep the logic together?

pshtt/pshtt.py
```diff
@@ -1134,30 +1134,46 @@ def load_suffix_list():
     return suffixes
 
 
-def initialize_external_data():
+def initialize_external_data(
+    init_preload_list=None,
```
IanLee1521: Personally I don't love the variable names, but I understand why you did them that way..

pshtt/pshtt.py
konklone: Yeah, I agree -- and I also think the cache interactions inside the `load_*` functions should move in there as well. The `load_*` functions can become simple functions that only do one thing, and the flag-processing code can just process the flag and get the third party cache dir stored for use in `initialize_external_data()`. I'll push an update to this PR moving that around.

konklone:
@IanLee1521 I refactored the code in 0456b73 to move all relevant caching logic into the one function. It's a more complicated function, but the other functions were simplified, and I removed 3 global variables from the library. Mind giving it a look?

konklone:
I should note, I also modified the signature of `load_suffix_list()` to return a tuple of `(suffix_list, content)` where `content` is the list of lines read from the file. This is necessary for any calling function to effectively cache the results.

konklone:
The Travis build breakages here look unrelated to this PR, but to an issue with `pytablereader` not being appropriately installed via `pip` during the Travis build prep process.

konklone:
After looking into this a bit more, I'd like to hold this open until I can make a change around the PSL. As it is, setting an instantiated PSL via Python won't work when data has to be serialized over the network, and adapting the function to take in a list of strings would be network-prohibitive in parallel, and sending in a partial PSL could result in unexpected behavior when evaluating redirects in the general case.

I think the right call is to provide a hook to disable the use of the PSL entirely, but I'm going to poke at it a bit more this week.

Feedback welcome on the best way to handle 1000 simultaneous pshtt processes in Lambda with a 200KB PSL, without either downloading 200MB of data from the PSL, or uploading 200MB to Lambda. :) 

I'm currently handling this for the preload list and preload pending list by doing the check locally and uploading a partial copy of the preload list (either a 0-element list or a 1-element list) and effectively monkeypatching pshtt to get the right answer with trivial time and network transfer.

One idea is to freeze a copy of the PSL into the container I upload to Lambda, which is replicated for every function execution. That might be acceptable, since while updates to the PSL are frequent enough, that's really more new gTLDs and the internet's long tail. For the function the PSL provides, it would be a lot more appropriate than freezing the preload list or pending list.

Anyway, I'll poke at this a bit more.

konklone:
OK, I added https://github.com/dhs-ncats/pshtt/pull/136/commits/b3061d1c25e9f6cda4e964cb3b87fff6e713b449 to make sure that caching the PSL via a .txt file is feasible.

I've also deployed this to Lambda and verified that the full pipeline works ([this is what I'm working with](https://github.com/18F/domain-scan/blob/eed8f8e9ad86ef3c9e8147d6c992459d61188827/scanners/pshtt.py) as of comment-time), at the cost of snapshotting the PSL inside my Lambda function environment. I can live with that, as the PSL snapshot will cover almost everything in a stable way, and I can now scan an arbitrary number of services with `pshtt` without incurring any significant network costs between my Lambda functions and third parties, or between my initiating server/computer and Lambda.

So I'm now :+1: for review and merge, if y'all are.

jsf9k:
I think the code I added in the `bugfix/missing_pytablereader` branch will fix the Travis build errors.

Also, I intend to get to reviewing this PR sometime today.

konklone:
Thanks for the reviews, all! This PR made parallelizing `pshtt` in the cloud far easier, I appreciate it.

PR #139, opened 2017-11-15
Added pytablereader to setup.py and requirements.txt

jsf9k:
It appears to get installed automatically with python 3 but not with python 2.  I'm not entirely certain what is going on, but see for example https://travis-ci.org/dhs-ncats/pshtt/jobs/301761188.

Also added test requirements to setup.py

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,7 @@ sslyze==1.1.0
 wget==3.2
 docopt
 requests_cache
+pytablereader
 pytablewriter
 publicsuffix
 pyopenssl==17.2.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,1 @@
-requests>=2.18.4
-sslyze==1.1.0
-wget>=3.2
-docopt
-pytablereader
-pytablewriter
-publicsuffix
-pyopenssl==17.2.0
-tox
-pytest
+-e .[test]
--- a/setup.py
+++ b/setup.py
@@ -63,10 +63,19 @@ setup(
         'wget>=3.2',
         'docopt',
         'requests_cache',
+        'pytablereader',
         'pytablewriter',
         'publicsuffix',
         'pyopenssl>=17.2.0'
     ],
+
+    extras_require={
+        # 'dev': ['check-manifest'],
+        'test': [
+            'tox',
+            'pytest'
+        ],
+    },
 
     # Conveniently allows one to run the CLI tool as `pshtt`
     entry_points={


jsf9k:
I'm wondering also what the use is in keeping a separate `requirements.txt` file in pshtt, since it doesn't contain any information not in `setup.py`.  I'd like to suggest removing it.  The [sample project from PyPA](https://github.com/pypa/sampleproject) does not include one.

IanLee1521:
@jsf9k -- My approach (not sure if it's the *right* one) is to use `setup.py` to define what the package _actually_ requires to run, while the `requirements.txt` is more of a developer resource. Therefore, it might include things like `coverage`, `pytest`, etc which wouldn't be needed to use the package, instead just to do development of it.

jsf9k:
@IanLee1521 -- `setup.py` does allow you to add extra requirements that are not installed unless you specifically request them.  See [here](https://github.com/pypa/sampleproject/blob/master/setup.py#L83-L90), for example.  That's what I did in this commit.

I'm still confused on the difference between `setup.py` and `requirements.txt` though.  It seems like you can put __concrete__ (package, version, __and__ where to get it from) dependencies in `requirements.txt` for repeatable builds, while `setup.py` can only contain __abstract__ dependencies (package and version __only__).  We aren't making use of concrete dependencies in the `pshtt` `requirements.txt` file, though.  I feel like we have two files that you have to keep in sync when you want to add, remove, or change a dependency, so that's why I'm bringing this up.

IanLee1521:
Right, the issue is that with putting the extra dependencies in `setup.py`, you have to install the package locally to get those. The requirements.txt is typically used in conjunction with a virtual environment, and then leverages the directory of Python files (in this case `pshtt-git-clone/pshtt/...`, rather than the files that the setup installation would place in your python lib path.

But I take your point... One thing that we could do to compress these might be to have the real requirements in `setup.py` and change the `requirements.txt` to something like:

```
# Install the package into the (virtual)-environment, in editable mode
-e .

# Extra testing stuff
coverage
pytest
ipython
...
```

jsf9k:
When I'm working on `pshtt` I work in a venv, and I just install the package to the local venv with `pip install -e .`.  With the changes I made to `setup.py` I can just run `pip install -e .[test]` instead or in addition if I want to run the tests.  I don't see the need for the separate `requirements.txt` file unless you want to use it to make concrete dependencies, which we're not doing.

How is the process you described different?

jsf9k:
I suppose we could also change the `requirements.txt` file to
```
-e .[test]
```

konklone:
I completely agree about it being a problem with these files getting out of sync. I think `requirements.txt` is nice to have around, and is often what people expect in a dev workflow.

It would also be required if we ever wanted to, even temporarily, pin to a git repository instead of a PyPi-published package. I'm currently doing that in `domain-scan` with `pshtt` itself, as `pshtt` is currently evolving faster than its PyPi publication rate.

The solution of using `-e .[test]` as the `requirements.txt` content to keep these in sync seems right to me, and also properly semantically expresses that a dev workflow requires installing the test dependencies in setup.py.

konklone:
Also, I just resolved the merge conflict (it was from me moving `requests_cache` in a previous PR).

PR #140, opened 2017-11-16
Fixed hanging on websites with never-ending streaming content

### LINKED ISSUE #138: Hanging on domains whose page load can never complete
`pshtt` hangs on domains like [lrostk.gsfc.nasa.gov](https://lrostk.gsfc.nasa.gov) and [gs-444-lrostk.gsfc.nasa.gov](https://gs-444-lrostk.gsfc.nasa.gov) where it can connect and bytes keep coming, but the page will never complete loading.  These particular pages load a single image of MIME type `multipart/x-mixed-replace`, which is a stream that never completes.

jsf9k:
This should resolve #138.

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -165,6 +165,13 @@ def ping(url, allow_redirects=False, verify=True):
     ONLY way the connection can be closed and released back into the
     pool.  One way to ensure this happens is to use the "with" Python
     construct.
+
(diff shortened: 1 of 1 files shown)


konklone:
Also, in my testing on the ~1300 federal domains, I didn't notice any time difference in the overall scan - it was 225 seconds vs 227 seconds, done with 900 Lambda workers.

PR #146, opened 2017-12-08
Flexible Dependencies for Development

egyptiankarim:
In addition to pinning dependencies in *master*, it might make sense to keep the *develop* branch nice and flexible, with the expectation that it might break.

I think it makes sense to retain the capability to build from "unpublished" source when necessary.

As a rule-of-thumb, maybe we also agree that the minimum version accepted by *develop* be the version we pin in *master*.


konklone:
What does this add to the release process of a new PyPi version? Do the version numbers have to change each time? Will they stay constantly out of sync between `develop` and `master`?

jsf9k:
@konklone, I think `develop` and `master` always being out of sync is a good point.

Better would be not to use the _current_ version for each package, but instead figure out the minimum version that will work.

konklone:
> Better would be not to use the current version for each package, but instead figure out the minimum version that will work.

This is mostly the case now:

```
    install_requires=[
        'requests>=2.18.4',
        'sslyze>=1.4.1',
        'wget>=3.2',
        'docopt',
        'pytablereader',
        'pytablewriter',
        'publicsuffix',
        'pyopenssl>=17.2.0'
    ],
```

Should we identify good minimum versions for the remaining 4 requirements and call it a day?

egyptiankarim:
> Should we identify good minimum versions for the remaining 4 requirements and call it a day?

I think that's the crux of this PR's principal commit 327ca184a51cfd0dccf2f3e9584c4a7d69ebf608.

PR #154, opened 2018-02-25
Redirects to www only matter if they're within the same hostname

### LINKED ISSUE #152: pshtt incorrectly labeling www as the canonical_endpoint when root redirects to *any* www URL
search.cio.gov is used to present search results for cio.gov.  If a valid search query is not provided, search.cio.gov redirects the user back to https://www.cio.gov/.
A *.cio.gov wildcard DNS record and certificate are used for search.cio.gov.  Due to the differences in wildcard handling between DNS and certificates, *.search.cio.gov subdomains will resolve, but HTTPS requests to them will result in a certificate error.
If we were expecting users to type this URL manually, we would likely add www.search.cio.gov as a SAN on the certificate to prevent that commonly-used prefix from resulting in a certificate error.  However, in this case, users are never expected to enter this URL directly, so we have no need to gracefully handle such quirks in user behavior.

When pshtt checks this domain, canonical_endpoint() detects the redirect being returned by search.cio.gov and tries to perform validation on www.search.cio.gov instead, which of course fails.  This causes search.cio.gov to incorrectly show up as non-compliant in Pulse.

konklone:
This updates the definition of `redirects_immediately_to_www` to only care about redirects to the corresponding `www` endpoint. 

Previously, that field was set if it redirected to any hostname beginning with `www`, including external redirects. However, the only reason we would ever care about redirecting to a hostname with `www` is to evaluate whether the domain appears to be redirecting within itself to favor the `www` hostname as its canonical endpoint.

The canonical endpoint function was adding an additional scoping around that to make sure it wasn't an external redirect, but that still didn't catch the case where it was redirecting to a `www` hostname within the same parent domain but not on the same evaluated hostname (e.g. `search.cio.gov` -> `www.cio.gov`).

This scopes the `redirects_immediately_to_www` field calculation, and drops the additional (now unneeded) scoping during canonical endpoint calculation.

There is an interesting tiny edge case where the `www` endpoint could itself redirect to another `www` prefix prepended, such as `https://www.example.gov` -> `https://www.www.example.gov`. This would set the `redirects_immediately_to_www` field for the `www` endpoint. That would be technically correct, and also wouldn't get factored into the calculation of the canonical endpoint for `example.gov` (which only interrogates the `http` and `https` endpoints), so I think that's fine.

Fixes #152.

cc @PaulSD

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -335,13 +335,20 @@ def basic_check(endpoint):
             base_immediate = parent_domain_for(subdomain_immediate)
 
             endpoint.redirect_immediately_to = immediate
-            endpoint.redirect_immediately_to_www = (re.match(r'^https?://www\.', immediate) is not None)
             endpoint.redirect_immediately_to_https = immediate.startswith("https://")
             endpoint.redirect_immediately_to_http = immediate.startswith("http://")
             endpoint.redirect_immediately_to_external = (base_original != base_immediate)
             endpoint.redirect_immediately_to_subdomain = (
                 (base_original == base_immediate) and
                 (subdomain_original != subdomain_immediate)
+            )
+
+            # We're interested in whether an endpoint redirects to the www version
+            # of itself (not whether it redirects to www prepended to any other
+            # hostname, even within the same parent domain).
+            endpoint.redirect_immediately_to_www = (
+                subdomain_immediate.startswith("www.") and
+                (re.sub("www\.", "", subdomain_immediate) == subdomain_original)
             )
 
             if ultimate_req is not None:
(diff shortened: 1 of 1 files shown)


jsf9k:
Nice catch, @konklone and @PaulSD!

pshtt/pshtt.py
```diff
@@ -344,6 +343,14 @@ def basic_check(endpoint):
                 (subdomain_original != subdomain_immediate)
             )
 
+            # We're interested in whether an endpoint redirects to the www version
+            # of itself (not whether it redirects to www prepended to any other
+            # hostname, even within the same parent domain).
+            endpoint.redirect_immediately_to_www = (
+                subdomain_immediate.startswith("www.") and
+                (re.sub("www\.", "", subdomain_immediate) == subdomain_original)
```
PaulSD: Would just `subdomain_intermediate == 'www.'+subdomain_original` be easier and more efficient here?

konklone: Yeah, probably! :) Went ahead and made this improvement (using interpolation) in #155.

PR #159, opened 2018-04-03
Upgrade sslyze to 1.4+ versions

jsf9k:


.travis.yml
```diff
@@ -3,7 +3,6 @@ language: python
 sudo: false
 
 python:
-  - '2.7'
```
konklone: I see the comment at https://github.com/nabla-c0d3/nassl/issues/34#issuecomment-375896876 that justifies this, and :+1:, but flagging this for watchers as a significant change -- this drops Python 2 support for `pshtt` entirely.

I think it would be worth (in a separate PR, which I can file) noting that this is a Python 3 only tool in the README.

PR #164, opened 2018-04-05
Autopublish to Docker Hub and PyPI when repository is tagged

jsf9k:
I now have TravisCI automagically deploying to Docker Hub and PyPI when the repository is tagged.

There is a lot of prerelease version bumping because that's how I was able to test the tagging+deployment automation.


konklone:
🎉 Nice job =)

PR #175, opened 2018-09-05
Upgrade to the latest sslyze

jsf9k:
The latest sslyze (2.0.0) no longer supports python versions before 3.6.


konklone (approved):
Looks great! Thanks for doing this.

PR #177, opened 2018-10-09
Sanity check test + dependency update

IanLee1521:
Attempts to catch any dependency issues with running `pshtt`

cc/ @egyptiankarim


jsf9k (approved):
Excellent catch!  I'm not sure why this is broken now.  Maybe one of the dependencies used to pull these in but no longer does?

PR #179, opened 2018-10-23
Add support for client authentication, some minor refactoring

konklone:
I'm opening this to more easily track @echudow's work on adding client cert authentication to `pshtt`.

There is some additional refactoring work in here, and possibly some logical changes. Let's use this PR to hash out the changes that would make `pshtt` usable for DHS, GSA, and DoD, without creating unintended regressions from existing reporting.

See also 18F/domain-scan#286.


echudow:
Because of the client cert issue, there could be some times when the initial request fails (and so it would not seem to be valid) even though it actually is valid so I changed one part where the initial request does not complete to assume that it is still valid and just couldn't complete the TLS handshake for other reasons, and then look at the sslyze results later to see if the HTTPS isn't valid.

echudow:
I also refactored some of the redirect code and I think fixed a bug there where it would report a redirect when there were no redirects.

echudow:
I also changed most of the log output to include the URL for easier troubleshooting based on the logs.

echudow:
@konklone Following up on the discussion from the other thread, are the results from my version actually harsher than the earlier `pshtt` version?  If so, which field?  It sounds like we're focusing on the Valid HTTPS field, but in the original version that field would only be true if the first request successfully completed, but the Python `requests` module I'm pretty sure throws an exception instead if the server cert is self-signed or has a bad chain (including missing Intermediate CA certs which is another issue), so the earlier version of `pshtt` should flag those as not valid.  If there's an exception, then `pshtt` tries again without cert validation but the `pshtt` logic did not mark those as valid https in that case.  Is this an issue in the original `pshtt` too or is there a different issue that I'm not seeing?

konklone:
@echudow My question about whether this was harsher on untrusted certs was a genuine question. :) I'm not sure, and have to go back and look at what the current outputs are.

echudow:
@konklone Ok, let me know what you find out and then I'll work through any issues so hopefully we can all use the same code in the end.

jsf9k:
@echudow and @konklone 
Sorry for letting this fall through the cracks.  Somehow I missed it.  I'm looking at it now.

jsf9k:
@echudow can you give me write permissions to your fork?  I want to merge the upstream changes into your branch, but I can't commit there.  The result of the merge is [here](https://github.com/dhs-ncats/pshtt/tree/improvement/merge_in_develop_from_upstream).  There were no conflicts.

The alternative is for me to create a pull request to your branch.

echudow:
Ok, I just gave you and Eric access to my forks of `pshtt` and `domain-scan`.

jsf9k:
Reviewing the code, I don't see any problems with these changes.  I could easily be missing something, but I think the only differences from the old code will be the new fields and sometimes the values of "liveness" and "is HTTPS valid".  Still, I need to do a full `pshtt` scan and compare with the old results.  I'm on vacation next week, but I can do that after I get back.

In the meantime, I added some reviewers in hope of getting more eyes on this pull request.  If folks want to test, I can also provide a few URLs that require client certificates:
* https://gotowork-piv-dr.ed.gov
* https://access.cpsc.gov
* https://access2.cpsc.gov
* https://workspace.usda.gov

IanLee1521:
I don't have much experience with testing / using client cert sites, but this functionally looks fine to me.  Do we have a verification that there aren't any regressions?

echudow:
Just fixed a bug that I introduced that prevented following some http redirects.  I'm sure that would have caused some issues with some incorrect data.  Has anyone compared the output to previous runs to see if there are any other bugs or regressions?  Hopefully it just adds some data instead of some earlier errors.

jsf9k:
@echudow I'm still planning to do that, but I keep getting derailed by other tasks.  My hope is to be able to do a full scan and check for regressions over the holidays.

pshtt/pshtt.py
```diff
@@ -217,75 +243,124 @@ def basic_check(endpoint):
...
+    # ip = req.raw._connection.sock.getpeername()
+    try:
+        if req.raw.closed is False:
+            ip = req.raw._connection.sock.socket.getpeername()[0]
+            if endpoint.ip is None:
+                endpoint.ip = ip
+            else:
+                if endpoint.ip != ip:
+                    utils.debug("{}: Endpoint IP is already {}, but requests IP is {}.".format(endpoint.url, endpoint.ip, ip))
+    except:
```
jsf9k: @echudow Can you catch the actual exception or exceptions you are expecting here?  A bare `except` is kind of scary, and `flake8` won't accept it.

Also, should we log something (at least at the debug level) if we end up in the exception clause?  Even if not, there should be a comment explaining why it's not a problem if we get an exception and why the `pass` is the right thing to do here.

jsf9k:
@echudow Once you remedy the bare `except` the TravisCI build should pass.

pshtt/pshtt.py
echudow: @jsf9k I should be able to make it catch an actual exception (even if it's just of type Exception).  I'm just trying to do best effort to get the actual IP address that is being scanned.  However, it looks like requests doesn't record the IP information so I have to reach down to the socket, but in some cases the socket is already closed and it throws an exception.  I try to limit the exceptions by checking if it's already closed, but even still it doesn't want a bare except statement.  I should be able to fix it later tonight.

jsf9k: Then just change it to catch `Exception` and add a comment with what you explained and I think it will be fine. :+1:

IanLee1521:
@jsf9k / @echudow -- Do you feel like this is ready to be merged? Looks like travis passed the PR build but hung on the branch build (sure...), I'm OK ignoring the coveralls drop which is the only real "fail" right now, though alternatively, is their more testing we could add for this to get the test coverage up?

Also, note that you are in a much better position than I am to confirm that this is working as expected and doesn't cause regressions.

If you're happy with it though, I can do the actual code review part for what I can comment on.

jsf9k:
@IanLee1521 and @echudow, this PR will break the BOD 18-01 scanning.  This is because [it treats some fields that were boolean as a union of a boolean and a string](https://github.com/cisagov/pshtt/blob/1ef217066fff6d9518ba66031aefc34f1df672ab/pshtt/pshtt.py#L170-L187).

This does give us more information than we had before, but I think it's important for the data fields to have a set type - particularly for the JSON output.  What if we by default assign fields a type of `None` and only set them to `True` or `False` if we actually know the result?  That will still break the BOD 18-01 scanning, and I will have to adjust it accordingly, but I think it's cleaner than using strings and booleans in the same field.

I can probably find some time to work on that whis weekend if you guys agree.

Thanks to @IanLee1521 for bumping this PR.  We need to get this approved and merged since it's practically old enough to vote at this point. :older_man:

echudow:
Thanks guys for looking at this again.  Leaving fields as `None` should work for us.  Before, `pshtt` would automatically turn almost all the fields that are `None` to `False`, but with the changes I made there are cases where there are partial scans and only some of the data is known and others are unknown.  Leaving them as `None` to indicate that they are unknown works for me and then we can convert those to "Unknown" when we put it up on our dashboard.

jsf9k:
This pull request **introduces 1 alert** when merging 52b9c7bd719c6a0da1502ce2e7c8d4824dbacc41 into 97c51ac3726642fac8198ef1a4b2ce32c8642f9e - [view on LGTM.com](https://lgtm.com/projects/g/cisagov/pshtt/rev/pr-a0a167d69f26cd17a9cf0ba8b29fa70e27f0e26b)

**new alerts:**

* 1 for Unnecessary pass

---

*Comment posted by [LGTM.com](https://lgtm.com)*

jsf9k:
This pull request **introduces 1 alert** when merging fbc318a25066ebfb08d817f600ab75bfa5986df7 into 97c51ac3726642fac8198ef1a4b2ce32c8642f9e - [view on LGTM.com](https://lgtm.com/projects/g/cisagov/pshtt/rev/pr-66f0cf504d202fdca00500b90d4722f496e0bb0f)

**new alerts:**

* 1 for Unnecessary pass

---

*Comment posted by [LGTM.com](https://lgtm.com)*

jsf9k:
This pull request **fixes 2 alerts** when merging f20eca483814ec5e442e387ad43a8128f151df33 into 97c51ac3726642fac8198ef1a4b2ce32c8642f9e - [view on LGTM.com](https://lgtm.com/projects/g/cisagov/pshtt/rev/pr-dd53b9743d6d39e1c504a71e6888681f84ddbae0)

**fixed alerts:**

* 2 for Unnecessary pass

---

*Comment posted by [LGTM.com](https://lgtm.com)*

jsf9k (approved):
Based on the results from doing a full test run with the new code, I think this PR is finally ready to be merged.

PR #180, opened 2018-11-07
Simplify "Enforces HTTPS" logic

jsf9k:
Based on the text in an HTTPS report he received, @DOS-cyber suggested that the logic
```
return is_domain_supports_https(domain) and (
    is_strictly_forces_https(domain) and
    (
        is_defaults_to_https(domain) or
        is_redirect(domain)
    ) or (
        (not is_strictly_forces_https(domain)) and
        is_defaults_to_https(domain)
    )
)
```
may be incorrect and should instead be:
```
return is_domain_supports_https(domain) and
    (
        is_strictly_forces_https(domain) and
        (
            is_defaults_to_https(domain) or
            is_redirect(domain)
        )
    ) or (
        (not is_strictly_forces_https(domain)) and
        is_defaults_to_https(domain)
    )
)
```
The logic is actually correct as it was originally written, because [`and` takes precedence over `or` in Python](https://docs.python.org/2.7/reference/expressions.html#operator-precedence).  Thus the parentheses are unnecessary.  This is admittedly confusing since it relies on an obscure implementation detail, and there is no way a report recipient could know the logic is implemented in Python, so I was going to make the suggested change here and in the HTTPS report text.

After staring at this bit of logic, though, I realized that if `is_defaults_to_https(domain)` is `True` then the value of `is_strictly_forces_https(domain)` is irrelevant.  This led me to 
realize that this expression can be simplified to:
```
return is_domain_supports_https(domain) and (
    is_defaults_to_https(domain) or (
        is_strictly_forces_https(domain) and is_redirect(domain)
    )
)
```

See also dhs-ncats/pshtt_reporter#23.


jsf9k:
The TravisCI build is currently failing due to `flake8` complaints.  I will make the necessary changes later to make the build pass.

DOS-cyber:
Thanks for the explanation and answering my question. When displayed in a pdf, it is plain English context and the syntax rules of Python don’t apply. However, it makes sense when you declare this language’s context. I’d still recommend parentheses in the plain English report that describes the methodology.

Great catch on reducing compute cycles with the simplified logic.

Neil

Neil Graves
Chief Cybersecurity Scientist
U.S. Department of State

________________________________

jsf9k:
Thanks for the issue report @DOS-cyber! I completely agree and intend to change the HTTPS report text as well. I'll tag you when I create that pull request tomorrow.

dav3r (approved):
Excellent work @jsf9k - it's nice to see this logic simplified 👏

konklone:
This looks right to me with my 👀. Have you checked whether there are any regressions between a scan before this change and after it?

jsf9k:
@konklone, I've done some spot checks and haven't turned up any differences.  I'm in the process right now of spinning up a staging environment to do a full scan run, but that probably won't actually happen until early next week.  (I don't want to mess up this weekend's production scan run.)

IanLee1521:
I'm not sure the size of the csv that pshtt / domain-scan generates, but I wonder if it would make sense to periodically commit the results of a full pshtt scan of dotgov into this repo?

We could then "snapshot" the state of all the websites into the tool repo, and use that to help track back regressions. My idealized assumption is that we would have minimal changes between runs (mostly just improvements in results, approaching a stable unchanged result) and that would help to keep down the size of the file in git space.

Thoughts on this approach?

konklone:
I suspect there'll be more flux than you might expect from a full scan of X0,000 hosts, from week to week. 

But this brings to mind the idea of creating a representative test suite of many situations that we could run automatically on each PR, which could help give us enough confidence about regressions that we'd have the confidence to make changes and channel any observed-in-the-wild regressions that evade our test suite, back into more tests. I think that's likely to be a more stable and complete approach.

jsf9k:
I did a full test run, and I don't see any regressions here.

PR #184, opened 2019-02-27
Org rename

jsf9k:


README.md
```diff
@@ -20,11 +20,11 @@ do](https://18f.gsa.gov/2017/01/04/tracking-the-us-governments-progress-on-movin
 
 `pshtt` is a collaboration between the Department of Homeland
```
h-m-f-t: I think that in most instances we can replace "Department of Homeland Security" with CISA (expanded or no). Yes, CISA is part of DHS, but no one says DHS TSA or DHS CBP. We're just CISA.

setup.py
```diff
@@ -28,10 +28,10 @@ def readme():
     # NCATS "homepage"
     url='https://www.dhs.gov/cyber-incident-response',
```
h-m-f-t: It's not great, but this page is probably the most representative page we have going for us right now:

https://www.us-cert.gov/resources/ncats

h-m-f-t (commented):
Small suggestions, but otherwise 👍

README.md
jsf9k: Fixed in 3f30e07.

setup.py
jsf9k: Fixed in e2c34d3.

h-m-f-t (approved):

PR #188, opened 2019-03-26
Use latest sslyze

jsf9k:
The latest version of sslyze has some bug fixes that we want to make sure we take advantage of.  (The latest version was most likely being installed anyway, but now it *must* be.)


dav3r (approved):
🎉

PR #189, opened 2019-03-31
Moved flake8 ignore logic from .travis.yml to setup.cfg

IanLee1521:
Simple update to move the `flake8` control flag logic out of travis and into setup.py so it effects everything equally.


jsf9k (approved):
:+1:

PR #193, opened 2019-05-17
Check for HSTS on all redirects on same endpoint

### LINKED ISSUE #89: HSTS Calculation not on final url
I mentioned some of the set up to this problem in the TTS #https-partner-support Slack channel, but the long and the short of the issue is that in https://github.com/dhs-ncats/pshtt/blob/00ff246f40acbea185d478d838c7fcd6652b9aa8/pshtt/pshtt.py#L80 the check is being done on "not the final https URL" as best I can tell (possibly it's actually randomly doing this...).

So for instance, as I'm looking at lc.llnl.gov, the redirects are: 

- http://lc.llnl.gov/
- https://lc.llnl.gov/
- https://lc.llnl.gov/nclogin.submit?LOGIN

And only the last hop (final URL) has the HSTS bits set up. This is the case for both lc.llnl.gov and lc-idm.llnl.gov, but for mylc.llnl.gov (which pshtt is currently saying is good for HSTS) the only difference I can find is in the middle hop:

- http://mylc.llnl.gov/
- https://lc.llnl.gov/lorenz
- https://lc.llnl.gov/nclogin.submit?LOGIN

Although, it also is not showing HSTS on the middle hop.

I'm still continuing to dig into this, but thought I would get the ticket opened while I work (since I still can't contribute directly yet).

echudow:
This is to resolve [**Issue #89 Factor in redirects when calculating HSTS**](https://github.com/cisagov/pshtt/issues/89) by looking at all redirects to the same endpoint for an HSTS header, rather than only the first response and no redirects.


jsf9k:
@echudow, I want to get #192 merged before merging this PR.  I am looking at the results of the full test run from that PR now.

pshtt/pshtt.py
```diff
@@ -525,6 +528,20 @@ def hsts_check(endpoint):
...
 
+        # If there is no HSTS header, check the eventual response from redirects
+        if header is None and endpoint.ultimate_req and endpoint.url in endpoint.ultimate_req.url:
+            header = endpoint.ultimate_req.headers.get("Strict-Transport-Security")
+            if header:
+                logging.warning("{}: Found HSTS in redirected response from {}.".format(endpoint.url, endpoint.ultimate_req.url))
+
+        # If there is still no HSTS header, check all the redirect responses
+        if header is None and endpoint.ultimate_req and endpoint.ultimate_req.history:
+            for entry in endpoint.ultimate_req.history:
```
jsf9k: I think the code here is assuming that the intermediate domains are somehow related to the initial domain.  What if the site redirects the user from `http://example.gov` to `http://www.example-2.gov` to `http://www.example-3.gov`?  If HSTS is present at the second or third sites, does that mean that we should give `http://example.gov` credit for being protected by HSTS?  It seems like the `http://example.gov` URL is still subject to hijack in that case, since the HSTS protection is associated with `example-2.gov` or `example-3.gov`.

Do we need to check a little deeper and make sure that the HSTS header would actually apply to the initial hostname?

I'll mark this as "needs changes" for now until we resolve this question.  I can undo that if I'm off the mark.

jsf9k (changes requested):
Please see my comment/question.

echudow:
@jsf9k, I added some code to record the HSTS URL and redirect chain so you can always see which URL the HSTS header that is being checking is coming from.  In the output there are now lines that say "<endpoint URL>: Found HSTS in redirected response from <redirected URL>".  In my tests, the code ensures that the redirected URL is on the same endpoint and I can see that in these lines in the output.

jsf9k:
This PR resolves #196.

jsf9k (approved):

pshtt/pshtt.py
```diff
@@ -521,9 +559,24 @@ def hsts_check(endpoint):
             return
```
jsf9k, line 559: It's not just a bad hostname that should cause the HSTS header to be ignored.  It is also ignored if the certificate is untrusted, for example.  (See item 2 in [the RFC](https://tools.ietf.org/html/rfc6797#section-2.2).)  Should this `if` clause be changed?

jsf9k (changes requested):
I think there are still a few open questions here that need to be answered.

pshtt/pshtt.py
jsf9k: I think this is still an open question.  I'd like to hear what @h-m-f-t thinks about this.

echudow: @jsf9k Based on @h-m-f-t 's comment at https://github.com/cisagov/pshtt/issues/89#issuecomment-318242757 saying 

> I don't want compliance for compliance' sake, or maximal-compliance for a little security benefit— and the current approach does have a bit of that flavor. Happy to hear more discussion, but I think I'm supportive of making this change, especially given that SSL Labs does this.

I think @h-m-f-t is already supportive of making this change.

pshtt/pshtt.py
echudow: @jsf9k Shouldn't that be in a separate issue/pull request to only give credit for HSTS if the endpoint's certificate is trusted?  Currently we don't check that the certificate is trusted even on the initial response, so that change could have more negative effects on the results.  This change here to check more responses from the same endpoint can only have positive results by giving some sites credit for HSTS that don't get credit for it today, so I think we should finish this one first and then make that change separately.

pshtt/pshtt.py
echudow: @jsf9k The check for `endpoint.url in redirect_entry.url` in the `if` https://github.com/cisagov/pshtt/blob/0ec75ba3fcabffa43b5c265d92c3404fc77e0a56/pshtt/pshtt.py#L571 already ensures that the HSTS header only counts if it is coming from the initial endpoint/hostname, so in your example `http://example.gov` would not get credit.

pshtt/pshtt.py
echudow: @jsf9k Based on @h-m-f-t 's comment at https://github.com/cisagov/pshtt/issues/89#issuecomment-318242757 saying 

> I don't want compliance for compliance' sake, or maximal-compliance for a little security benefit— and the current approach does have a bit of that flavor. Happy to hear more discussion, but I think I'm supportive of making this change, especially given that SSL Labs does this.

I think @h-m-f-t is already supportive of making this change.

jsf9k: > Shouldn't that be in a separate issue/pull request to only give credit for HSTS if the endpoint's certificate is trusted?

You're right.  I created a separate issue for that.  See #205.

pshtt/pshtt.py
jsf9k: > @jsf9k The check for `endpoint.url in redirect_entry.url` in the `if`
> 
> https://github.com/cisagov/pshtt/blob/0ec75ba3fcabffa43b5c265d92c3404fc77e0a56/pshtt/pshtt.py#L571
> 
> already ensures that the HSTS header only counts if it is coming from the initial endpoint/hostname, so in your example `http://example.gov` would not get credit.

It gives credit if _any_ HSTS header encountered in the redirect chain involves the initial endpoint/hostname.  If the redirect chain goes on to send the user to an HTTP URL on an entirely different domain for which the browser has never seen an HSTS header, then it seems to me that the user is not being protected by the HSTS header.

Do you think this is a valid concern?  I'm OK deferring the resolution to #205, or a separate issue.  I feel like there is a hole here, and if so then I'd like to at least document that before moving on.  Once we put this last little issue to bed I think I'm OK with approving.

pshtt/pshtt.py
jsf9k: > @jsf9k Based on @h-m-f-t 's comment at [#89 (comment)](https://github.com/cisagov/pshtt/issues/89#issuecomment-318242757) saying
> 
> > I don't want compliance for compliance' sake, or maximal-compliance for a little security benefit— and the current approach does have a bit of that flavor. Happy to hear more discussion, but I think I'm supportive of making this change, especially given that SSL Labs does this.
> 
> I think @h-m-f-t is already supportive of making this change.

Understood, but I'd like him to be more explicit about it by approving this pull request.  In any event, we need two non-author approvals to merge it.

pshtt/pshtt.py
echudow: I think it is a valid concern, but I think it is a separate issue.  I think that the requirement for HSTS is only for the domain that we're checking, not for all endpoints/domains in the redirect chain.  Today we would give credit for HSTS to a domain if it returns a valid HSTS header in the initial response even if it then immediately redirects to an HTTP URL on an entirely different domain (although it should fail the separate Strictly Forces HTTPS check).  Here we are just continuing what we already do. 

I also think it is separate from https://github.com/cisagov/pshtt/issues/205 too. BOD 18-01 requires 
> All publicly accessible Federal websites and web services provide service through a secure connection

so I think it should fail some bigger check for a secure connection.  We already have a separate check for Domain Enforces HTTPS, so I think this check should be implemented there instead to check that there are no downgrades anywhere in the redirect chains for the domain's endpoints.  We could even add an additional check called Strictly Enforces HSTS to check that all endpoints in the redirect chain also have HSTS if you and the CyHy team think that should be a requirement.

jsf9k: Created #206.

jsf9k (approved):

jsf9k:
Closed due to a lack of attention.

Issue #195, opened 2019-06-20
Other requests response exception


jsf9k:
# 🐛 Bug Report

When performing pshtt scanning, sometimes we get an error in AWS Lambda stating "other requests response exception".  In this case we do not receive scan results for the domain that errored.

## To Reproduce

`caats3.va.gov` is a domain where this happens often.

## Expected behavior

A successful scan.

## Any helpful log output

None.

echudow:
@jsf9k I wonder if this has to do with the breaking change in `sslyze` v2.1.0 in early June where they changed how certificate verification works and so a number of the fields in the certificate plugin went away.  I think there were some missing attribute exceptions.  See the commits in my fork at https://github.com/echudow/pshtt/commit/c2f427b005f799ea7e916a5b067dbb4c55cce3f9 and https://github.com/echudow/pshtt/commit/ce0ccaf39c51aa7b5d97e14ef04c0fd4f1cc7240 for how I addressed this issue with some workarounds to handle both the older and newer versions of sslyze.  We're still testing those changes though.

jsf9k:
@echudow This issue can't be due to that, since the pshtt lambda hadn't been rebundled since May.  Thanks for mentioning that, though, since it will be a problem now (as I'm rebundling to deploy the logging improvements in #197, which will pull in a newer sslyze).

PR #197, opened 2019-07-30
Improve exception logging

jsf9k:
These changes should help debug #195 and the issue seen this past weekend where the "valid HTTPS" field was `None` in some cases due to the sslyze certificate check throwing an unexpected exception.


echudow (approved):

dav3r (approved):
Log 'em all! ⚡️

pshtt/pshtt.py
```diff
@@ -602,16 +599,24 @@ def https_check(endpoint):
...
+            if "timed out" in str(err):
+                logging.exception("{}: Retrying sslyze scanner certificate plugin.".format(endpoint.url))
                 cert_plugin_result = scanner.run_scan_command(server_info, command)
+            else:
+                logging.exception("{}: Unknown exception in sslyze scanner certificate plugin.".format(endpoint.url))
+                utils.debug("{}: {}".format(endpoint.url, err))
+                endpoint.unknown_error = True
+                # We could make this False, but there was an error so
+                # we don't know
+                endpoint.https_valid = None
```
konklone, line 611: Instead of picking `True` or `False` when there's an unknown exception, I suggest having the scan error out and report no data for this domain. It doesn't need to crash the entire scanning process, but I think it would be healthier for error classes that the `pshtt` maintainers haven't yet encountered to create some user-visible and dev-visible friction that incentivizes tracking down the new exception class and handling it appropriately.

I also think that from a customer perspective, it's better for `pshtt` to go from Nothing to True/False than to go from True to True/False as it gets more precise over time.

echudow: @konklone, Is setting `unknown_error` to `True` sufficient to error out, or is there more that is needed?

I completely agree that it's better to return Nothing if we don't know for sure rather than True which could then later turn into a False.

PR #203, opened 2019-08-13
Update HSTS check

mcdonnnj:
This PR would update the `hsts_check()` function's method of processing the HSTS header. Currently we do not closely conform to RFCs 7230 and 6797 with regard to valid header formats. This is specifically related to domains returning multiple `Strict-Transport-Security` headers. This has allowed some domains to pass checks they might not otherwise because they are responding with an invalid HSTS header format.

This change would result in domains that previously passed some checks now failing them. However, if it is a result of them failing to conform to the RFC specification then I believe that is appropriate. This is still a lenient change because we use the first header instead of immediately failing if multiple headers are seen.

This is in part a result of the requests library assuming that a response with multiple header fields with the same field name conform to 7230-3.2.2. It concatenates them with a comma so in a server incorrectly outputting multiple `Strict-Transport-Security` headers it combines them following the guidelines of 7230-3.2.2.

Example HTTP response (just `Strict-Transport-Security` fields from curl output):
```
< Strict-Transport-Security: max-age=31536000
< Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
< Strict-Transport-Security:  max-age=31536000
```
becomes
`"max-age=31536000, max-age=31536000; includeSubDomains; preload, max-age=31536000"`

The same domain run against the Qualys SSL Labs SSL Server Test:
>Server sent invalid HSTS policy. See below for further information.
>Strict Transport Security (HSTS) | Invalid   Server provided more than one HSTS header

Per RFC 7230 section 3.2.2:
> A sender MUST NOT generate multiple header fields with the same field
   name in a message unless either the entire field value for that
   header field is defined as a comma-separated list [i.e., #(values)]
   or the header field is a well-known exception (as noted below).

Per RFC 6797 section 6.1:
```
Strict-Transport-Security = "Strict-Transport-Security" ":"
                            [ directive ]  *( ";" [ directive ] )

directive                 = directive-name [ "=" directive-value ]
directive-name            = token
directive-value           = token | quoted-string

where:
     token          = <token, defined in [RFC2616], Section 2.2>
     quoted-string  = <quoted-string, defined in [RFC2616], Section 2.2>
```


pshtt/pshtt.py
```diff
@@ -526,31 +526,40 @@ def hsts_check(endpoint):
...
+        # Multiple HSTS headers does not conform to RFCs 7230-3.2.2 and 6797-6.1
+        if len(headers) > 1:
+            logging.warning("Host is incorrectly returning multiple HSTS headers: {}".format(header))
+
+        # Put all of the directives in the HSTS header into a dictionary
+        directive_list = [x.strip() for x in headers[0].split(";")]
+        directives = dict()
+        for directive in directive_list:
+            components = directive.split("=")
+            directives[components[0]] = "".join(components[1:]) or True
```
jsf9k: This line probably deserves a comment explaining what it does.  It's relying on the fact that the empty string evaluates to `False` and the fact that the `or` operator short circuits without forcing the whole result to be cast to a bool.

Instead of a comment, it might be clearer to break it up something like this:
```python
key = components[0]
value = "".join(components[1:])
# If the value is the empty string then there was no "=" present, and hence
# the key corresponds to a boolean flag.
if value == "":
    value = True
directives[key] = value
```

I'm fine with either way, as long as it's made clearer what is going on.

pshtt/pshtt.py
```diff
@@ -526,31 +526,40 @@ def hsts_check(endpoint):
...
+            return
 
-        temp = re.split(r';\s?', second_pass)
+        endpoint.hsts = True
 
-        if "max-age" in header.lower():
-            endpoint.hsts_max_age = int(temp[0][len("max-age="):])
+        endpoint.hsts_max_age = int(directives["max-age"]) if "max-age" in directives else None
 
         if endpoint.hsts_max_age is None or endpoint.hsts_max_age <= 0:
```
jsf9k: Are these values disallowed according to the RFC?  If so we should add a comment stating that.

pshtt/pshtt.py
```diff
@@ -526,31 +526,40 @@ def hsts_check(endpoint):
...
-        # Set max age to the string after max-age
-        # TODO: make this more resilient to pathological HSTS headers.
-
         # handle multiple HSTS headers, requests comma-separates them
-        first_pass = re.split(r',\s?', header)[0]
-        second_pass = re.sub(r'\'', '', first_pass)
+        headers = [x.strip() for x in header.split(",")]
+        # Multiple HSTS headers does not conform to RFCs 7230-3.2.2 and 6797-6.1
+        if len(headers) > 1:
+            logging.warning("Host is incorrectly returning multiple HSTS headers: {}".format(header))
```
jsf9k, line 538: We could just fail them here, if multiple headers are not allowed according to the RFC.  Did you choose not to for a technical reason, or just to avoid making sweeping changes?

If we can point folks to a section of the RFC for support then we can make bolder changes that negatively impact scoring.  It's when the change is in the "interpretive" logic in pshtt_reporter that evaluates things like "Supports HTTPS" that we should be very careful.

jsf9k (changes requested):
I had a few questions, and there are a few lines that I think merit comments.

pshtt/pshtt.py
mcdonnnj: This was definitely a sweeping changes moment. I thought it more conservative to give credit for the first header received vs failing completely. I am quite open to failing them completely on multiple headers (which appears to be along the lines of what SSL Labs does). However, since browsers do not die on it that may be a harder point to get across.

pshtt/pshtt.py
mcdonnnj: Added a comment explaining what kind of value we're expecting to help explain why we check that.

pshtt/pshtt.py
mcdonnnj: I wasn't happy with how it looked after coming back to it so I just reworked it to make it clear what's going on.

pshtt/pshtt.py
jsf9k: This is a point to discuss with the CyHy team, after explaining the situation to them.  I'd in particular want input from @climber-girl, and @h-m-f-t if he is available.

pshtt/pshtt.py
```diff
@@ -526,31 +526,50 @@ def hsts_check(endpoint):
...
+        endpoint.hsts_max_age = int(directives["max-age"]) if "max-age" in directives else None
 
+        # According to section 6.1.1 of the HSTS RFC, a non-negative
+        # integer value is required.  A zero value indicates that the
+        # user agent should cease considering the host a known HSTS
+        # host, and therefore only strictly positive values indicate
+        # the presence of HSTS protection.  See
+        # https://tools.ietf.org/html/rfc6797#section-6.1.1 for more
+        # details.
         if endpoint.hsts_max_age is None or endpoint.hsts_max_age <= 0:
```
jsf9k: I think there is an argument here for letting `endpoint.hsts` indicate only the presence of a syntactically correct HSTS header.  That value can be combined with `endpoint.hsts_max_age` and `endpoint.hsts_all_subdomains` to determine if the domain "supports HSTS."  To that end, I think it makes sense to change `endpoint.hsts_max_age <= 0` to `endpoint.hsts_max_age < 0` since a max-age value of zero is valid according to [the RFC](https://tools.ietf.org/html/rfc6797#section-6.1.1).

What do you think, @mcdonnnj?

jsf9k (changes requested):
We still have an open question regarding how we handle multiple headers, and I added a further question related to the max-age value evaluation.

pshtt/pshtt.py
mcdonnnj: I agree on changing the `<=` to just `<` and only checking for syntactical correctness in this part of the code.

mcdonnnj: I agree on changing `<=` to `<` and only checking for syntactical correctness here. That makes sense for what this part of the code is trying to do (which is just establish that we're getting a valid header and what it contains from the response).

pshtt/pshtt.py
mcdonnnj: After digging in the RFC more deeply it does specify in [Section 8.1](https://tools.ietf.org/html/rfc6797#section-8.1)

>If a UA receives more than one STS header field in an HTTP response message over secure transport, then the UA MUST process only the first such header field.

So although sending multiple headers is wrong, the RFC does have guidelines for how to handle such cases.

climber-girl: I agree with @mcdonnnj on changing the code to not immediately fail for multiple headers and rather take the first header.

pshtt/pshtt.py
climber-girl: @jsf9k @mcdonnnj I understand wanting to impart valid HSTS syntax, but rather than saying HSTS=True when the max-age is set to zero, could we instead Fail "HSTS" while still reporting the HSTS header in "HSTS Header" since technically the a value of zero means "forget me as an HSTS host/I don't do HSTS"?  Alternatively, we could relabel the "HSTS" column as "Valid HSTS" if you'd prefer that?

jsf9k: @climber-girl A valid of zero _is_ valid HSTS, according to the RFC.

Don't forget that pshtt is not the end of the line when it comes to the BOD reporting.  The pshtt results are interpreted via cisagov/pshtt_reporter, and that is where the actual report is created.  In this case, they would still fail "Supports HSTS" in the report, since max-age is too small.

In the past we have tried to make pshtt and trustymail simply _collect_ information that is then _interpreted_ by the reporting code.

climber-girl: @jsf9k I understand that it is valid per the RFC, hence the secondary alternative I proposed to relabel the current HSTS column that is reported (which would I guess be included in the pshtt_reporter?). For the purposes of collecting info with pshtt, I'm fine with the change you and @mcdonnnj discussed.

echudow: I already thought that the HSTS column was referring to whether it was valid since it can already be `False` and still have data in the HSTS Header column.  I don't think the column name needs to be changed since that is already how it is used and there may be downstream effects from changing the column name.  Generally, though, I like the `trustymail` way of having one column for existence and then another column for validity.

pshtt/pshtt.py
jsf9k: It seems that the consensus is to generate the warning for multiple headers and take the first one.  I'm OK with that.

pshtt/pshtt.py
jsf9k: @climber-girl, if the value is zero and the `HSTS` column is `True`, then the `Domain Uses Strong HSTS` column will still be `False`.  Therefore the domain will "fail HSTS" in the HTTPS report.  The `HSTS` column in the raw pshtt results only says whether a valid HSTS header is being served, but the `Domain Uses Strong HSTS` includes the max-age check.

PR #204, opened 2019-08-15
WIP:  Changed is_hsts() function to check both https endpoints.

### LINKED ISSUE #194: HSTS not checked at both https endpoints
While looking at egeo.usss.gov for HTTPS/HSTS issues per customer request for assistance to understand why 'Enforces HTTPS' was failing (based on the most recent code change we made here: https://github.com/cisagov/pshtt/pull/192, http-www endpoint was correctly causing this failure because it doesn't redirect to an https site), I noticed that the https-www endpoint also does not have an HSTS header.  The report is incorrectly showing Strong HSTS = True because it is based on the HSTS header of the plain https endpoint.

I think a change similar to the "Bugfix for domain_enforces_https() logic" at link above needs to be done to ensure that both the plain https and the https-www endpoints are checked for Strong HSTS.

$ curl --head egeo.usss.gov
HTTP/1.1 301 Moved Permanently
Content-Length: 145
Content-Type: text/html; charset=UTF-8
Location: https://egeo.usss.gov/
Server: Microsoft-IIS/10.0
X-Powered-By: ASP.NET
Date: Fri, 07 Jun 2019 15:38:14 GMT

$ curl --head https://egeo.usss.gov
HTTP/2 200 
content-length: 783
content-type: text/html
last-modified: Tue, 14 May 2019 19:52:48 GMT
accept-ranges: bytes
etag: "1d50a8e9b092b0f"
server: Kestrel
set-cookie: ApiUrl=https%3A%2F%2Fegeo.usss.gov%2FAtlasLogin%2F; path=/
x-powered-by: ASP.NET
strict-transport-security: max-age=31536000
date: Fri, 07 Jun 2019 15:38:26 GMT

$ curl --head www.egeo.usss.gov
HTTP/1.1 404 Not Found
Content-Length: 315
Content-Type: text/html; charset=us-ascii
Server: Microsoft-HTTPAPI/2.0
Date: Fri, 07 Jun 2019 15:38:53 GMT
Connection: close

$ curl --head https://www.egeo.usss.gov
HTTP/2 404 
content-length: 315
content-type: text/html; charset=us-ascii
server: Microsoft-HTTPAPI/2.0
date: Fri, 07 Jun 2019 15:38:41 GMT

mcdonnnj:
Work-in-progress:  DO NOT MERGE!!

This PR is to fix #194 by checking both https and httpswww endpoints to determine HSTS status for the domain.

### DIFF
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -1206,11 +1206,14 @@ def is_hsts(domain):
     if not https.live and not httpswww.live:
         return None
 
-    hsts = True
-    if https.live:
-        hsts &= https.hsts
-    if httpswww.live:
-        hsts &= httpswww.hsts
+    hsts = None
+    if https.live and (https.hsts is not None):
+        hsts = https.hsts
+    if httpswww.live and (httpswww.hsts is not None):
+        if hsts is None:
+            hsts = httpswww.hsts
+        else:
+            hsts &= httpswww.hsts
 
     return hsts
 
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -1199,16 +1199,20 @@ def is_missing_intermediate_cert(domain):
 
 def is_hsts(domain):
     """
-    Domain has HSTS if its canonical HTTPS endpoint has HSTS.
+    Domain has HSTS if both https and httpswww endpoints have HSTS when live.
     """
-    canonical, https, httpswww = domain.canonical, domain.https, domain.httpswww
+    https, httpswww = domain.https, domain.httpswww
 
-    if canonical.host == "www":
-        canonical_https = httpswww
-    else:
-        canonical_https = https
+    if not https.live and not httpswww.live:
+        return False
 
-    return canonical_https.hsts
+    hsts = True
+    if https.live:
+        hsts &= https.hsts
+    if httpswww.live:
+        hsts &= httpswww.hsts
+
+    return hsts
 
 
 def hsts_header(domain):
--- a/pshtt/pshtt.py
+++ b/pshtt/pshtt.py
@@ -1204,7 +1204,7 @@ def is_hsts(domain):
     https, httpswww = domain.https, domain.httpswww
 
     if not https.live and not httpswww.live:
-        return False
+        return None
 
     hsts = True
     if https.live:


pshtt/pshtt.py
```diff
@@ -1199,16 +1199,20 @@ def is_missing_intermediate_cert(domain):
 
 def is_hsts(domain):
     """
-    Domain has HSTS if its canonical HTTPS endpoint has HSTS.
+    Domain has HSTS if both https and httpswww endpoints have HSTS when live.
```
jsf9k, line 1202: The canonical endpoint was (I think) introduced by @h-m-f-t in an effort to make the HTTPS report results more digestible by the less technical.  I'd have to know his thoughts on this change before approving.

mcdonnnj: I think @climber-girl 's point was that if both endpoints are live then both should have HSTS. I think [this section of RFC6797](https://tools.ietf.org/html/rfc6797#section-11.4.2) touches on what I mean although it is in the context of a web application and the includeSubDomains directive.

echudow (dismissed):
LGTM, just remember that this has the potential to negatively effect the results for a number of domains, so be prepared for questions.

dav3r (approved):
This solution works for me- the docstring is clear and the code is clean. 👍

pshtt/pshtt.py
echudow: I don't completely understand all of the reasons for only focusing on the canonical endpoint. It seems like we assume that that is where most users will end up, but plenty of others may use the other endpoint (if it is live), and attackers could take advantage of the other endpoint (via spear-phishing URLs and the like) if it is not sufficiently protected. 

`strictly_forces_https` already looks at both http endpoints, not just the canonical one, but `downgrades_https` currently only looks at the canonical https endpoint. If we switch this HSTS check to check both endpoints, then we should probably change some other checks to look at both endpoints (such as `downgrades_https` and maybe even follow-on `sslyze` checks).  Should most of these checks for both endpoints be in new "Strictly_Enforces" checks or should these be changed in the existing checks?

echudow:
I committed changes to address when the hsts field is None.  There might be a more concise way to do this using Python logic, but I was very confused at first seeing that in Python when there are None's the order of an `or` matters so that `(None or False) != (False or None)`.

mcdonnnj:
@echudow in `hsts_check()` it explicitly sets `endpoint.hsts = False` if there is no STS header. Would you elaborate on what circumstances you would see a `None` for the `endpoint.hsts` value?

echudow:
@mcdonnnj You're right. Sorry, I was testing this code change in my fork which is a little different and leaves `endpoint.hsts` as `None` if it couldn't fully connect to the site and get the headers.  Your original code should work here.  My update should work too even without there ever being any None values since that is a subcase (and it would be easier for me to merge in the upstream changes).

mcdonnnj:
@echudow Ah, that makes sense. I agree that there's no harm in leaving the change. I just wanted to make sure it was needed and I see why it's useful for compatibility with your fork.

PR #267, opened 2026-06-16
Honor explicit ports in HTTPS checks

### LINKED ISSUE #254: Ports other than 443 result in null values where there shouldn't be null values
# 🐛 Summary #

"HTTPS Custom Truststore Trusted" and "HTTPS Cert Chain Length" check results are `null` when specifying a port number with a URL. These should not be `null`, and should instead be representative of the actual values reflected.

## To reproduce ##

- Scan a normal website without a port number with a custom CA file: `pshtt --timeout 2 --ca-file=custom_ca_file.pem google.com`
- Observe that these fields are not null
- Scan the same website, but with the port number: `pshtt --timeout 2 --ca-file=custom_ca_file.pem google.com:443`
- Observe that these fields are null
- Scan any other website on any port specified: `pshtt --timeout 2 --ca-file=custom_ca_file.pem example.com:9443`
- Observe that these fields are null

## Expected behavior ##

These should not be `null`, and should instead be representative of the actual values reflected. SSLyze, for example, gets these values correct even when the port is provided, but PSHTT does not.

## Any helpful log output or screenshots ##

Paste the results here:

```console
arcsector@device ~/test pshtt> pshtt --timeout 2 --ca-file=google_root.pem google.com:443 --debug --json 2> /dev/null

-------------------------

Fetching Chrome preload list from source...
Starting new HTTPS connection (1): chromium.googlesource.com:443
https://chromium.googlesource.com:443 "GET /chromium/src/+/main/net/http/transport_security_state_static.json?format=TEXT HTTP/1.1" 200 None

-------------------------

Fetching hstspreload.org pending list...
Starting new HTTPS connection (1): hstspreload.org:443
https://hstspreload.org:443 "GET /api/v2/pending HTTP/1.1" 200 None

-------------------------

Pinging http://google.com:443...
Starting new HTTP connection (1): google.com:443
[
  {
    "Base Domain": "google.com:443",
    "Base Domain HSTS Preloaded": false,
    "Canonical URL": "https://www.google.com:443",
    "Defaults to HTTPS": true,
    "Domain": "google.com:443",
    "Domain Enforces HTTPS": true,
    "Domain Supports HTTPS": true,
    "Domain Uses Strong HSTS": false,
    "Downgrades HTTPS": false,
    "HSTS": false,
    "HSTS Entire Domain": false,
    "HSTS Header": null,
    "HSTS Max Age": null,
    "HSTS Preload Pending": false,
    "HSTS Preload Ready": false,
    "HSTS Preloaded": false,
    "HTTPS Bad Chain": false,
    "HTTPS Bad Hostname": false,
    "HTTPS Cert Chain Length": null, # <--- HERE
    "HTTPS Client Auth Required": false,
    "HTTPS Custom Truststore Trusted": null, # <--- HERE
    "HTTPS Expired Cert": false,
    "HTTPS Full Connection": true,
    "HTTPS Live": true,
    "HTTPS Probably Missing Intermediate Cert": false,
    "HTTPS Publicly Trusted": null,
    "HTTPS Self Signed Cert": false,
    "IP": null,
    "Live": true,
    "Notes": "",
    "Redirect": false,
    "Redirect To": null,
    "Server Header": "gws",
    "Server Version": null,
    "Strictly Forces HTTPS": true,
    "Unknown Error": true,
    "Valid HTTPS": true
    ...
  }
]
```

arpitjain099:
Fixes #254

This came from https_check always building the SSLyze target with port 443 after slicing the URL string. If someone scans `domain:9443`, the port was not handled correctly and cert trust or chain fields ended up null.

This updates target parsing to use `urlparse` and pass the parsed host plus parsed port to SSLyze. When no explicit port is present, it still defaults to 443.

I also added unit tests that assert calls to `with_ip_address_lookup` for:
- `example.com:9443` -> host `example.com`, port `9443`
- `example.com` -> host `example.com`, port `443`

I could not run `pytest` end-to-end in this local environment because the `nassl` binary fails to load (`_SSLv2_method` symbol issue on macOS toolchain). I still verified changes by compiling modified files and running a targeted Python harness with stubbed SSLyze modules to validate host and port parsing.

### DIFF
--- a/src/pshtt/_version.py
+++ b/src/pshtt/_version.py
@@ -1,3 +1,3 @@
 """This file defines the version of this module."""
 
-__version__ = "0.7.3"
+__version__ = "0.7.4"
--- a/src/pshtt/pshtt.py
+++ b/src/pshtt/pshtt.py
@@ -654,12 +654,21 @@ def https_check(endpoint):
     """Use sslyze to figure out the reason an endpoint failed to verify."""
     utils.debug("sslyzing %s...", endpoint.url)
 
-    # remove the https:// from prefix for sslyze
+    # Parse endpoint URL so explicit non-default ports are preserved.
     try:
-        hostname = endpoint.url[8:]
+        parsed_url = urlparse.urlparse(endpoint.url)
+        hostname = parsed_url.hostname
+        port = parsed_url.port
+
+        if hostname is None:
+            raise ValueError(f"Unable to parse hostname from URL: {endpoint.url}")
+
+        if port is None:
+            port = 443 if parsed_url.scheme == "https" else 80
+
         server_location = (
             ServerNetworkLocationViaDirectConnection.with_ip_address_lookup(
-                hostname=hostname, port=443
+                hostname=hostname, port=port
             )
         )
         server_tester = ServerConnectivityTester()
--- a/tests/test_pshtt.py
+++ b/tests/test_pshtt.py
@@ -101,7 +101,9 @@ class TestHttpsCheckServerLocation(unittest.TestCase):
 
         server_location = MagicMock()
         server_location.ip_address = "127.0.0.1"
-        mock_server_network_location.with_ip_address_lookup.return_value = server_location
+        mock_server_network_location.with_ip_address_lookup.return_value = (
+            server_location
+        )
 
         tester = MagicMock()
         tester.perform.side_effect = RuntimeError("stop after checking location args")
@@ -123,7 +125,9 @@ class TestHttpsCheckServerLocation(unittest.TestCase):
 
         server_location = MagicMock()
         server_location.ip_address = "127.0.0.1"
-        mock_server_network_location.with_ip_address_lookup.return_value = server_location
+        mock_server_network_location.with_ip_address_lookup.return_value = (
+            server_location
+        )
 
         tester = MagicMock()
         tester.perform.side_effect = RuntimeError("stop after checking location args")
--- a/tests/test_pshtt.py
+++ b/tests/test_pshtt.py
@@ -109,7 +109,12 @@ class TestHttpsCheckServerLocation(unittest.TestCase):
         tester.perform.side_effect = RuntimeError("stop after checking location args")
         mock_server_connectivity_tester.return_value = tester
 
-        https_check(endpoint)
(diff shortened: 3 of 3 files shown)


jsf9k (approved):

jsf9k:
@arpitjain099 - Are the [stack traces that appear](https://github.com/cisagov/pshtt/actions/runs/27696084364/job/81920008730?pr=267#step:8:62) when running the tests you added necessary?  Is there any way to get rid of them?  They misled me into thinking that tests had failed when I ran `pytest` locally.

jsf9k:
> @arpitjain099 - Are the [stack traces that appear](https://github.com/cisagov/pshtt/actions/runs/27696084364/job/81920008730?pr=267#step:8:62) when running the tests you added necessary? Is there any way to get rid of them? They misled me into thinking that tests had failed when I ran `pytest` locally.

Hmmm...looks like the stack traces are indeed necessary because of the way that `pshtt` handles exceptions.

arpitjain099:
Good catch, and they're not actually necessary - that was the test leaking pshtt's own exception logging. Both tests stop the scan early by raising from `perform()`, and `https_check` catches that and logs it with `logging.exception`, so the traceback printed even though the tests passed. I wrapped those calls in `with self.assertLogs(level="ERROR"):` so the expected error log is captured instead of printed. Pushed - output should be quiet now.

dav3r (approved):
This all looks good, but I'd like to see one more test added: `test_http_check_defaults_to_port_80` since that logic is being added in this PR as well.

dav3r:
> This all looks good, but I'd like to see one more test added: `test_http_check_defaults_to_port_80` since that logic is being added in this PR as well.

Doh- I didn't notice that this PR had auto-merge enabled, otherwise I would've used the "Request Changes" option.

arpitjain099:
> > This all looks good, but I'd like to see one more test added: `test_http_check_defaults_to_port_80` since that logic is being added in this PR as well.
> 
> 
> 
> Doh- I didn't notice that this PR had auto-merge enabled, otherwise I would've used the "Request Changes" option.

@dav3r no worries let me check

arpitjain099:
@dav3r opened #268 to add `test_http_check_defaults_to_port_80`. Thanks!

PR #268, opened 2026-06-23
Add test for HTTP default port 80 in https_check

arpitjain099:
Follow-up to #267. @dav3r noted that #267 also added the `http` -> port 80 default but only the port 443 and explicit-port cases had tests, and it auto-merged before that could be addressed. This adds `test_http_check_defaults_to_port_80` to cover the remaining branch.

It mirrors the existing `test_https_check_defaults_to_port_443` test, just with an http endpoint and port 80.

Note: I couldn't run the suite locally (sslyze's nassl native lib won't load on my macOS), but I verified the logic directly: `Endpoint("http", "root", "example.com").url` is `http://example.com`, and the port parse in `https_check` returns 80 for an http URL with no explicit port. CI should exercise the full test on Linux like its siblings.

### DIFF
--- a/tests/test_pshtt.py
+++ b/tests/test_pshtt.py
@@ -148,3 +148,32 @@ class TestHttpsCheckServerLocation(unittest.TestCase):
         mock_server_network_location.with_ip_address_lookup.assert_called_once_with(
             hostname="example.com", port=443
         )
+
+    @patch("pshtt.pshtt.ServerConnectivityTester")
+    @patch("pshtt.pshtt.ServerNetworkLocationViaDirectConnection")
+    def test_http_check_defaults_to_port_80(
+        self, mock_server_network_location, mock_server_connectivity_tester
+    ):
+        """Default to port 80 when an http endpoint URL has no explicit port."""
+        endpoint = Endpoint("http", "root", "example.com")
+
+        server_location = MagicMock()
+        server_location.ip_address = "127.0.0.1"
+        mock_server_network_location.with_ip_address_lookup.return_value = (
+            server_location
+        )
+
+        tester = MagicMock()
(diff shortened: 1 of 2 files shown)


jsf9k:
@arpitjain099 - Please let me know what you think of commits [d659b65](https://github.com/cisagov/pshtt/pull/268/commits/d659b656f1fae24abee6f2eee6ae83cfc87a48af) and [01f6a13](https://github.com/cisagov/pshtt/pull/268/commits/01f6a13215d684b52fdbeb73d711c0bc157a7440).

jsf9k (approved):

tests/test_pshtt.py
```diff
@@ -120,13 +123,24 @@ def test_https_check_uses_explicit_port(
...
+        [
+            param(80, "http", id="http"),
+            param(443, "https", id="https"),
+        ],
+    )
     @patch("pshtt.pshtt.ServerConnectivityTester")
     @patch("pshtt.pshtt.ServerNetworkLocationViaDirectConnection")
-    def test_https_check_defaults_to_port_443(
-        self, mock_server_network_location, mock_server_connectivity_tester
+    def test_https_check_default_ports(
```
dav3r: The current test name is a little wonky now that this test checks both http and https.  Not sure if this is better, but I took a whack at something:
```suggestion
    def test_check_default_web_ports(
```

dav3r (approved):
LGTM, aside from the test name comment I mentioned.  👍

arpitjain099:
Looks great @jsf9k - folding the two into a parametrized `test_check_default_web_ports` is cleaner than my separate test, and it picks up @dav3r's naming suggestion too. The unittest-parametrize approach and the mypy ignore for it both look good to me. Thanks for tightening it up.

PR #270, opened 2026-07-08
fix: coerce is_strictly_forces_https return value to bool

### LINKED ISSUE #176: "Strictly Forces HTTPS" can sometimes be null
Hi,

I've noticed that for some sites I'm scanning, the Strictly Forces HTTPS field is sometimes `null`. I suspect this is for similar reasons as #49.

Here's a representative example:

```
pshtt --json https://projects.houstonchronicle.com/newstips/
```
results in (some fields snipped for brevity):
```
[
  {
    "Live": true,
    "Redirect": false,
    "Redirect To": null,
    "Strictly Forces HTTPS": null,
    "Unknown Error": false,
    "Valid HTTPS": true,
    "endpoints": {
      "http": {
        "live": true,
        "redirect": null,
        "redirect_immediately_to_external": null,
        "redirect_immediately_to_http": null,
        "redirect_immediately_to_https": null,
        "redirect_immediately_to_subdomain": null,
        "redirect_immediately_to_www": null,
        "status": 200,
        "unknown_error": false,
        "url": "http://projects.houstonchronicle.com/newstips/"
      },
      "https": {
        "live": true,
        "redirect": null,
        "redirect_immediately_to": null,
        "redirect_immediately_to_external": null,
        "redirect_immediately_to_http": null,
        "redirect_immediately_to_https": null,
        "redirect_immediately_to_subdomain": null,
        "redirect_immediately_to_www": null,
        "status": 200,
        "unknown_error": false,
        "url": "https://projects.houstonchronicle.com/newstips/"
      },
      "httpswww": {
        "live": false,
        "redirect": null,
        "redirect_immediately_to": null,
        "redirect_immediately_to_external": null,
        "redirect_immediately_to_http": null,
        "redirect_immediately_to_https": null,
        "redirect_immediately_to_subdomain": null,
        "redirect_immediately_to_www": null,
        "status": null,
        "unknown_error": false,
        "url": "https://www.projects.houstonchronicle.com/newstips/"
      },
      "httpwww": {
        "headers": {},
        "live": false,
        "redirect": null,
        "redirect_immediately_to": null,
        "redirect_immediately_to_external": null,
        "redirect_immediately_to_http": null,
        "redirect_immediately_to_https": null,
        "redirect_immediately_to_subdomain": null,
        "redirect_immediately_to_www": null,
        "status": null,
        "unknown_error": false,
        "url": "http://www.projects.houstonchronicle.com/newstips/"
      }
    }
  }
]
```

My theory is that this is due to not being strict (no pun intended) with booleans in [the code here](https://github.com/dhs-ncats/pshtt/blob/develop/pshtt/pshtt.py#L796), where we have

```
def down_or_redirects(endpoint):
        return ((not endpoint.live) or endpoint.redirect_immediately_to_https)
```

which will return `None` if `live = True` and `redirect_immediately_to_https = None`, since this evaluates to `False or None`, which is itself `None`. Very similar to #49.

Anyway, thanks for reading. Hope this all makes sense. There is a comment on the other ticket about wanting all fields to be either true or false, so I think this issue could be considered a bug under that definition.

steps-re:
Fixes #176.

## Problem

`is_strictly_forces_https` could return `None` instead of `True`/`False`, which caused the "Strictly Forces HTTPS" JSON field to be `null` for some domains.

The root cause was in the `down_or_redirects` inner helper:

```python
# before
def down_or_redirects(endpoint):
    return not endpoint.live or endpoint.redirect_immediately_to_https
```

`redirect_immediately_to_https` defaults to `None` in `models.py` (line 58). When `endpoint.live` is `True`, Python evaluates `False or None` as `None`. That `None` then propagates through the `and` chain on line 1280, so the final return value is `None` rather than a bool.

## Fix

Wrap the attribute in `bool()` in the helper, and apply the same treatment to `https_somewhere`:

```python
def down_or_redirects(endpoint):
    return (not endpoint.live) or bool(endpoint.redirect_immediately_to_https)

https_somewhere = bool(https.live or httpswww.live)
```

Both `bool(None)` → `False` and `bool(True)` → `True`, so the semantics are unchanged for any non-None value.

## Tests

Adds `TestStrictlyForcesHttps` with three cases:
- All endpoints at default (None) state → returns `False`, not `None`
- HTTPS live, HTTP live with `redirect_immediately_to_https=None` → returns `False`, not `None`
- HTTPS live, HTTP endpoints both redirecting to HTTPS → returns `True`

### DIFF
--- a/src/pshtt/_version.py
+++ b/src/pshtt/_version.py
@@ -1,3 +1,3 @@
 """This file defines the version of this module."""
 
-__version__ = "0.7.5"
+__version__ = "0.7.6"
--- a/src/pshtt/pshtt.py
+++ b/src/pshtt/pshtt.py
@@ -1274,9 +1274,9 @@ def is_strictly_forces_https(domain):
     )
 
     def down_or_redirects(endpoint):
-        return not endpoint.live or endpoint.redirect_immediately_to_https
+        return (not endpoint.live) or bool(endpoint.redirect_immediately_to_https)
 
-    https_somewhere = https.live or httpswww.live
+    https_somewhere = bool(https.live or httpswww.live)
     all_http_unused = down_or_redirects(http) and down_or_redirects(httpwww)
 
     return https_somewhere and all_http_unused
--- a/tests/test_pshtt.py
+++ b/tests/test_pshtt.py
@@ -10,7 +10,12 @@ from unittest_parametrize import ParametrizedTestCase, param, parametrize
 
 # cisagov Libraries
 from pshtt.models import Domain, Endpoint
-from pshtt.pshtt import certificate_is_expired, https_check, is_live
+from pshtt.pshtt import (
+    certificate_is_expired,
+    https_check,
+    is_live,
+    is_strictly_forces_https,
+)
 
 
 class TestLiveliness(unittest.TestCase):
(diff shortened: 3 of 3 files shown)


jsf9k (approved):

jsf9k:
Thank you for the contribution @steps-re!

dav3r (approved):
LGTM - thanks @steps-re for your contribution! 👍

